PR #1, opened 2020-12-18
REFACTOR: Convert to single reaction per user

Ahmedgagan:
Draft PR

### DIFF
--- a/app/services/discourse_reactions/reaction_manager.rb
+++ b/app/services/discourse_reactions/reaction_manager.rb
@@ -14,7 +14,7 @@ module DiscourseReactions
       ActiveRecord::Base.transaction do
         return if (@like && !@guardian.can_delete_post_action?(@like)) || (is_reacted_by_user && !@guardian.can_delete_reaction_user?(old_reacted_user))
         @reaction = reaction_scope&.first_or_create
-        @reaction_user = reaction_user_scope&.first_or_initialize
+        @reaction_user = reaction_user_scope
         @reaction_value == DiscourseReactions::Reaction.main_reaction_id ? toggle_like : toggle_reaction
       end
     end
@@ -55,8 +55,8 @@ module DiscourseReactions
     def reaction_user_scope
       return nil unless @reaction
       search_reaction_user = DiscourseReactions::ReactionUser.where(user_id: @user.id, post_id: @post.id)
-      create_reaction_user = DiscourseReactions::ReactionUser.where(reaction_id: @reaction.id, user_id: @user.id, post_id: @post.id)
-      search_reaction_user.length > 0 ? search_reaction_user : create_reaction_user
+      create_reaction_user = DiscourseReactions::ReactionUser.new(reaction_id: @reaction.id, user_id: @user.id, post_id: @post.id)
+      search_reaction_user.length > 0 ? search_reaction_user.first : create_reaction_user
     end
 
     def is_reacted_by_user
@@ -73,11 +73,15 @@ module DiscourseReactions
 
     def remove_shadow_like
       PostActionDestroyer.new(@user, @post, post_action_like_type).perform
+      delete_like_reaction
+    end
+
+    def delete_like_reaction
       DiscourseReactions::Reaction.where("reaction_value = '#{DiscourseReactions::Reaction.main_reaction_id}' AND post_id = ?", @post.id).destroy_all
     end
 
     def add_reaction
-      @reaction_user = reaction_user_scope&.first_or_initialize unless is_reacted_by_user
+      @reaction_user = reaction_user_scope unless is_reacted_by_user
       @reaction_user.save!
       add_reaction_notification
     end
@@ -85,6 +89,10 @@ module DiscourseReactions
     def remove_reaction
       @reaction_user.destroy
       remove_reaction_notification
+      delete_reaction
+    end
+
+    def delete_reaction
       DiscourseReactions::Reaction.where("reaction_users_count = 0 AND post_id = ?", @post.id).destroy_all
     end
 
--- a/app/services/discourse_reactions/reaction_manager.rb
+++ b/app/services/discourse_reactions/reaction_manager.rb
@@ -7,12 +7,14 @@ module DiscourseReactions
       @user = user
       @guardian = guardian
       @post = post
-      @reaction = reaction_scope.first_or_create
       @like = @post.post_actions.find_by(user: @user, post_action_type_id: post_action_like_type)
     end
 
     def toggle!
       ActiveRecord::Base.transaction do
+        return if (@like && !@guardian.can_delete_post_action?(@like)) || (is_reacted_by_user && !@guardian.can_delete_reaction_user?(old_reacted_user))
+        @reaction = reaction_scope&.first_or_create
+        @reaction_user = reaction_user_scope&.first_or_initialize
         @reaction_value == DiscourseReactions::Reaction.main_reaction_id ? toggle_like : toggle_reaction
       end
     end
@@ -20,7 +22,6 @@ module DiscourseReactions
     private
 
     def toggle_like
-      raise Discourse::InvalidAccess if @like && !@guardian.can_delete_post_action?(like)
       remove_shadow_like if @like
       remove_reaction if is_reacted_by_user
       add_shadow_like unless @like
@@ -53,11 +54,17 @@ module DiscourseReactions
 
     def reaction_user_scope
       return nil unless @reaction
-      DiscourseReactions::ReactionUser.where(reaction_id: @reaction.id, user_id: @user.id, post_id: @post.id)
+      search_reaction_user = DiscourseReactions::ReactionUser.where(user_id: @user.id, post_id: @post.id)
+      create_reaction_user = DiscourseReactions::ReactionUser.where(reaction_id: @reaction.id, user_id: @user.id, post_id: @post.id)
+      search_reaction_user.length > 0 ? search_reaction_user : create_reaction_user
     end
 
     def is_reacted_by_user
       DiscourseReactions::ReactionUser.find_by(user_id: @user.id, post_id: @post.id) ? true : false
+    end
+
+    def old_reacted_user
+      DiscourseReactions::ReactionUser.find_by(user_id: @user.id, post_id: @post.id)
     end
 
     def add_shadow_like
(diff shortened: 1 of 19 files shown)


app/services/discourse_reactions/reaction_manager.rb
```diff
@@ -50,7 +54,17 @@ def reaction_scope
 
     def reaction_user_scope
       return nil unless @reaction
-      DiscourseReactions::ReactionUser.where(reaction_id: @reaction.id, user_id: @user.id)
+      search_reaction_user = DiscourseReactions::ReactionUser.where(user_id: @user.id, post_id: @post.id)
+      create_reaction_user = DiscourseReactions::ReactionUser.where(reaction_id: @reaction.id, user_id: @user.id, post_id: @post.id)
+      search_reaction_user.length > 0 ? search_reaction_user : create_reaction_user
```
jjaffeux: do we do the create query everytime even if sometimes we dont need it here?

app/services/discourse_reactions/reaction_manager.rb
```diff
@@ -50,7 +54,17 @@ def reaction_scope
...
+    end
+
+    def is_reacted_by_user
+      DiscourseReactions::ReactionUser.find_by(user_id: @user.id, post_id: @post.id) ? true : false
```
jjaffeux: do we need the true/false ? it will return nil or something here anyways no?

app/services/discourse_reactions/reaction_manager.rb
```diff
@@ -59,18 +73,20 @@ def add_shadow_like
 
     def remove_shadow_like
       PostActionDestroyer.new(@user, @post, post_action_like_type).perform
+      DiscourseReactions::Reaction.where("reaction_value = 'heart' AND post_id = ?", @post.id).destroy_all
```
jjaffeux: can we use some constant for heart here or dont we have a site setting?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -339,8 +339,18 @@ export default createWidget("discourse-reactions-actions", {
     this.state.statePanelExpanded = false;
     this.state.reactionsPickerExpanded = true;
     this.scheduleRerender();
+
+    const currentUserReactions = this.attrs.post.current_user_reactions;
+    let currentUsedReaction;
+
+    if (currentUserReactions[0] && currentUserReactions[0].id !== "heart") {
```
jjaffeux: same thing about heart

jjaffeux (changes requested):
I see you are working around a lot the fact that se might have multiple reactions and doing things like reactions[0], we don't want that.

Having one reaction is how it works now, the code should reflect it. Why do we have reactionS everywhere if I can only have one?

app/services/discourse_reactions/reaction_manager.rb
Ahmedgagan: yes, we always use`first_or_initialize` method with this method

plugin.rb
```diff
@@ -81,33 +82,33 @@ class Engine < ::Rails::Engine
...
     like = object.post_actions.find do |l|
       l.post_action_type_id == PostActionType.types[:like] &&
       l.deleted_at.blank? &&
       l.user_id == scope.user.id
     end
 
-    return reactions if like.blank?
+    return nil if like.blank?
+
     like_reaction = {
```
jjaffeux, line 105: what is it used for now?

plugin.rb
```diff
@@ -81,33 +82,33 @@ class Engine < ::Rails::Engine
...
+      return {
         id: reaction.reaction_value,
         type: reaction.reaction_type.to_sym,
         can_undo: reaction_user.can_undo?
-      }
-    end.compact
+      } if reaction.reaction_users_count
+    end
 
     like = object.post_actions.find do |l|
```
jjaffeux, line 101: we could just have this as the last statement here, if it's not nil we will return this, otherwise nil

plugin.rb
```diff
@@ -57,7 +57,8 @@ class Engine < ::Rails::Engine
   end
 
   add_to_serializer(:post, :reactions) do
-    reactions = object.reactions.map do |reaction|
+
+    reactions = object.reactions.reject { |hash| hash[:reaction_users_count] == nil }.map do |reaction|
```
jjaffeux: do we need to check on nil here ?

plugin.rb
Ahmedgagan: If there's no reaction and no like, then it will return nil rather than going further

assets/stylesheets/common/discourse-reactions.scss
```diff
@@ -247,12 +247,19 @@ html.discourse-reactions-no-select {
...
   touch-action: none;
 
-  .btn-toggle-reaction {
+  .btn-toggle-reaction-like {
     outline: none;
     pointer-events: none;
     background: none !important;
   }
 
+  img.btn-toggle-reaction-emoji {
```
jjaffeux: do we need to style on "img" here ? it's adding specificity

plugin.rb
Ahmedgagan: Yes, as we are taking like as a reaction, we take an entry in the table with reaction_users_count = nil, so we avoid that reaction here

app/services/discourse_reactions/reaction_manager.rb
jjaffeux: What I mean is that create_reaction_user query will get executed each time, even if `search_reaction_user.length` > 0 given we call it before, we should avoid it, it's unnecessary work

Ahmedgagan: Yes that's right, ill try doing it in other way. As when we create the user we need `reaction_id` to create user successfully

plugin.rb
jjaffeux: No I mean ruby will do this without == nil no ?

Ahmedgagan: That's right 😅

plugin.rb
Ahmedgagan: here we are using `each` loop, we don't need to return any value. I am returning the value as we have only one reaction per user and if we find any reaction we directly return this value to withdraw control from this method.

plugin.rb
```diff
@@ -57,7 +57,8 @@ class Engine < ::Rails::Engine
   end
 
   add_to_serializer(:post, :reactions) do
-    reactions = object.reactions.map do |reaction|
+
+    reactions = object.reactions.reject { |reaction| !reaction[:reaction_users_count] }.map do |reaction|
```
jjaffeux: instead of rejecting what "it's not", maybe you could keep what "it is"?

jjaffeux:
Im surprised we disnt have to change tests, could you check we are not missing some ruby specs here please?

Easiest thing to do is xommenting one line you changed and ran the specs, if nothing breaks... well 😅

Don't go for 100% coverage but look for specific bits of logic, conditionals for example...

app/services/discourse_reactions/reaction_manager.rb
```diff
@@ -7,27 +7,31 @@ def initialize(reaction_value:, user:, guardian:, post:)
...
+      remove_reaction if is_reacted_by_user
+      add_shadow_like unless @like
     end
 
     def toggle_reaction
       PostAction.limit_action!(@user, @post, post_action_like_type)
-      @reaction = reaction_scope.first_or_create
-      @reaction_user = reaction_user_scope&.first_or_initialize
-      @reaction_user.persisted? ? remove_reaction : add_reaction
+      remove_reaction if is_reacted_by_user
```
jjaffeux, line 33: I would do:

```ruby
if is_reacted_by_user
  remove_reaction
else
  add_reaction
end
```

Ahmedgagan: in this case, if the user wants to change its reaction, we need to first remove the reaction and then add a new one

app/services/discourse_reactions/reaction_manager.rb
```diff
@@ -7,27 +7,31 @@ def initialize(reaction_value:, user:, guardian:, post:)
...
       end
     end
 
     private
 
     def toggle_like
-      like = @post.post_actions.find_by(user: @user, post_action_type_id: post_action_like_type)
-      raise Discourse::InvalidAccess if like && !@guardian.can_delete_post_action?(like)
-      like ? remove_shadow_like : add_shadow_like
+      remove_shadow_like if @like
```
jjaffeux, line 25:
```ruby
if @like
  remove_shadow_like
else
  add_shadow_like
end
```

assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js.es6
```diff
@@ -60,17 +75,45 @@ export default createWidget("discourse-reactions-reaction-button", {
       title = "discourse_reactions.main_reaction.cant_remove";
     }
 
-    return { title: I18n.t(title) };
+    if (
+      currentUserReaction &&
+      currentUserReaction.can_undo &&
+      !likeAction.hasOwnProperty("can_undo")
```
jjaffeux, line 84: This is not something we do a lot in the codebase, it would be worth a comment to explain why you decided to use hasOwnProperty here

plugin.rb
```diff
@@ -81,33 +84,33 @@ class Engine < ::Rails::Engine
     reactions << like_reaction
   end
 
-  add_to_serializer(:post, :current_user_reactions) do
-    return [] unless scope.user.present?
-    reactions = object.reactions.map do |reaction|
+  add_to_serializer(:post, :current_user_reaction) do
+    return nil unless scope.user.present?
```
jjaffeux, line 88: for example this line here should have caused a test change, we are probably missing serializers tests

app/services/discourse_reactions/reaction_manager.rb
jjaffeux: so you mean, we count on the fact that between both calls the result of is_reacted_by_user  will have changed?

Ok, well that's slightly error prone, but let's go with this for now

app/services/discourse_reactions/reaction_manager.rb
Ahmedgagan: it is the same as the case of the reaction, if the user wants to change its like, we need to first remove the remove and then add a new one

plugin.rb
Ahmedgagan: This test is available in the code, i have fixed in the latest commit

Ahmedgagan:
> Im surprised we disnt have to change tests, could you check we are not missing some ruby specs here please?
> 
> Easiest thing to do is xommenting one line you changed and ran the specs, if nothing breaks... well 😅
> 
> Don't go for 100% coverage but look for specific bits of logic, conditionals for example...

there were some tests breaking, in the latest commit i have fixed all the tests

jjaffeux:
Please look at the remaining comments then we will merge it

app/services/discourse_reactions/reaction_manager.rb
Ahmedgagan: I also got the same of changing is_reacted_by_user, added a fix in the latest commit

Ahmedgagan:
Added a commit with all suggested changes.

assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js.es6
```diff
@@ -75,6 +75,7 @@ export default createWidget("discourse-reactions-reaction-button", {
       title = "discourse_reactions.main_reaction.cant_remove";
     }
 
+    // !likeAction.hasOwnProperty("can_undo")  it specifies user have not liked the post
```
jjaffeux: no that's not the comment I was expected :) why do you do:

```
!likeAction.hasOwnProperty("can_undo")
```

and not

```
!likeAction.can_undo
```

?

This is what you should explain

Ahmedgagan: how about this?
hasOwnProperty("can_undo") can_undo property is included only if the post is liked by user. So, it is used here to check whether users have not liked the post or not irrespective of the can_undo value which can be true or false. Liked posts can't be assigned the title = "discourse_reactions.picker.remove_reaction" OR "discourse_reactions.picker.cant_remove_reaction".

Ahmedgagan: OR this
used `!likeAction.hasOwnProperty("can_undo")` rather than `!likeAction.can_undo` to check whether can_undo property is present or not irrsepective of its value

jjaffeux (approved):

PR #3, opened 2021-01-06
UX: suppressing summary panel

Ahmedgagan:
suppressing summary panel when there's only one reaction and the current user has the used the same reaction

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-list.js.es6
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-list.js.es6
@@ -8,8 +8,13 @@ export default createWidget("discourse-reactions-list", {
 
   html(attrs) {
     const reactions = attrs.post.reactions;
+    const currentUserReaction = attrs.post.current_user_reaction;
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/widgets/discourse-reactions-list.js.es6
```diff
@@ -8,11 +8,16 @@ export default createWidget("discourse-reactions-list", {
 
   html(attrs) {
     const reactions = attrs.post.reactions;
+    const currentUserReaction = attrs.post.current_user_reaction;
 
     if (attrs.post.reaction_users_count <= 0) {
       return;
     }
 
+    if(currentUserReaction && reactions.length == 1 && reactions[0].id == currentUserReaction.id) {
```
jjaffeux: is it not enough to just check for currentUserReaction and reactions.length === 1 ?

How could it be different if there's only one reaction and we know current user has reacted?

jjaffeux (approved):

PR #8, opened 2021-01-28
FIX: Fixed the Animation of reaction-button

Ahmedgagan:


assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -132,8 +132,8 @@ export default createWidget("discourse-reactions-actions", {
     }
 
     if (
-      attrs.post.reactions.length == 1 &&
-      attrs.post.reactions[0].id == "heart"
+      attrs.post.reactions.length === 1 &&
+      attrs.post.reactions[0].id === "heart"
```
jjaffeux: use constant please

assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js.es6
```diff
@@ -18,17 +18,21 @@ export default createWidget("discourse-reactions-reaction-button", {
...
-          postId: this.attrs.post.id,
-          canUndo: currentUserReaction.can_undo
-        });
-      } else {
-        this.callWidgetFunction("toggleLike");
-      }
+      this.callWidgetFunction("toggleLike", {
+        reaction: currentUserReaction ? currentUserReaction.id : null
+      });
+      // if (hasUsedMainReaction) {
```
jjaffeux: why do we keep this ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -251,19 +251,17 @@ export default createWidget("discourse-reactions-actions", {
               this.setCurrentUserReaction();
               later(() => {
```
jjaffeux: do we still need this later ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -345,59 +347,97 @@ export default createWidget("discourse-reactions-actions", {
...
+            },
+            attrs,
+            true
+          );
         });
       });
     });
   },
 
+  toggleCustomReaction(
```
jjaffeux: really not liking this function, this is a monster, can you try to use composition to make it more understandable ?

Ahmedgagan: It's my fault 😅 I did this change for other reason, but that didn't work and I forgot to remove it

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: Yes, later is still needed here

assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js.es6
Ahmedgagan: Sorry for this😅

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: can you elaborate why ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -345,54 +355,69 @@ export default createWidget("discourse-reactions-actions", {
...
+          let toggleReaction =
+            attrs.reaction && attrs.reaction !== mainReactionName
+              ? attrs.reaction
+              : this.siteSettings.discourse_reactions_reaction_for_like;
+
+          CustomReaction.toggle(this.attrs.post.id, toggleReaction).then(
+            value => {
+              resolve();
+
+              if (value === undefined) {
```
jjaffeux: what are the possible values of value here ?

assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js.es6
```diff
@@ -18,17 +18,9 @@ export default createWidget("discourse-reactions-reaction-button", {
...
-      } else if (currentUserReaction) {
-        this.callWidgetFunction("toggleReaction", {
-          reaction: currentUserReaction.id,
-          postId: this.attrs.post.id,
-          canUndo: currentUserReaction.can_undo
-        });
-      } else {
-        this.callWidgetFunction("toggleLike");
-      }
+      this.callWidgetFunction("toggleLike", {
```
jjaffeux: I don't get why we have to pass the reaction to toggleLike? toggleLike should be responsible for this logic.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -345,54 +355,69 @@ export default createWidget("discourse-reactions-actions", {
...
+                const mainReactionIcon = this.siteSettings
+                  .discourse_reactions_like_icon;
+                const hasUsedMainReaction = this.attrs.post
+                  .current_user_used_main_reaction;
+                const template = document.createElement("template");
+                template.innerHTML = iconHTML(
```
jjaffeux: extract this logic please

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -345,54 +355,69 @@ export default createWidget("discourse-reactions-actions", {
...
-              this.attrs.post.current_user_used_main_reaction = current_user_used_main_reaction;
-              this.attrs.post.reactions = reactions;
-              this.attrs.post.reaction_users_count = reaction_users_count;
-              this.scheduleRerender();
+          let toggleReaction =
+            attrs.reaction && attrs.reaction !== mainReactionName
+              ? attrs.reaction
+              : this.siteSettings.discourse_reactions_reaction_for_like;
+
+          CustomReaction.toggle(this.attrs.post.id, toggleReaction).then(
```
jjaffeux: why don't we do : `CustomReaction.toggle(....).then(resolve).then(...)` ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -345,54 +355,69 @@ export default createWidget("discourse-reactions-actions", {
           .current_user_used_main_reaction;
         const template = document.createElement("template");
         template.innerHTML = iconHTML(
-          hasUsedMainReaction ? `far-${mainReactionIcon}` : mainReactionIcon
+          hasUsedMainReaction ||
```
jjaffeux, line 355: extract this please

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -345,54 +355,69 @@ export default createWidget("discourse-reactions-actions", {
...
+          } else {
+            this.addUserReaction(
+              this.siteSettings.discourse_reactions_like_icon
+            );
+          }
 
-          if (!this.attrs.post.current_user_reaction) {
+          if (this.attrs.post.current_user_reaction) {
+            this.attrs.post.reaction_users_count -= 1;
+            this.setCurrentUserReaction();
```
jjaffeux: I find it a little bit weird that we use setCurrentUserReaction with no arg, we don't set anything, we actually unset

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: Either Undefined OR json response

assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js.es6
Ahmedgagan: Oooh sorry, i forgot to change the method name😅 we have used this method as it doesn't performs reaction animation

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: so why not: `if (!value)` ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: also if it's needed please look at what I do with later in others parts of the code, we can let a later uncontrolled like this

Ahmedgagan: If we don't use it later here, the `drop animation` stops working, This is due to the `this.scheduleRerender()` it rerenders the page which eventually stops the animation.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: when there's the ERROR in API, it returns undefined where `! value` doesn't works

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: We can change the method's name. Do you have any suggestions?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: Capture d’écran 2021-01-29 à 15 09 11

I don't undertand

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: unsetCurrentUserReaction ?

Ahmedgagan: or we can just modify the name of the current method as `setOrUnsetCurrentUserReaction`?

jjaffeux: I think the easiest is to keep setCurrentUserReaction() but pass it null setCurrentUserReaction(null) in this case, it will make it more explicit.

Ahmedgagan: Sure, that's too will look nice

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: I did a complete check for this, it was not needed here. did a change in the new commit.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -250,23 +250,20 @@ export default createWidget("discourse-reactions-actions", {
...
-                    params.reaction
-                  ).then(value => {
-                    resolve();
+                  return CustomReaction.toggle(params.postId, params.reaction)
+                    .then(resolve)
+                    .catch(e => {
+                      bootbox.alert(`${e.jqXHR.status} ${e.errorThrown}`);
 
-                    if (value === undefined) {
                       this.attrs.post.current_user_reaction = current_user_reaction;
```
jjaffeux, line 262: why is it in the catch now ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -291,19 +288,17 @@ export default createWidget("discourse-reactions-actions", {
...
-                    if (value === undefined) {
-                      this.attrs.post.current_user_reaction = current_user_reaction;
-                      this.attrs.post.current_user_used_main_reaction = current_user_used_main_reaction;
-                      this.attrs.post.reactions = reactions;
-                      this.attrs.post.reaction_users_count = reaction_users_count;
-                      this.scheduleRerender();
-                    }
-                  }
-                );
+                    this.attrs.post.current_user_reaction = current_user_reaction;
```
jjaffeux, line 297: same why in catch ?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: extract in a function outside of this scope, I want to reduce complexity of the whole function please

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: It was written because, when there's some error in API, we can undo the change we made in the frontend. So, I replaced `then` with `catch` which now works perfectly.

Also, I don't know the reason why but this condition was always true, and due to this, it was causing some issues in UI which was realized now when I testing this condition.

jjaffeux: please dig a better explanation :)

Ahmedgagan: The reason we were using `then` here was, when there's some error in the backend we can undo the changes we did from the frontend. Eg: we add a 🙂 reaction, but there's 500 error. So in this situation, our frontend is updated with 🙂 reaction. To undo this we were using `then` but that was not meant for that use. So, I replaced it with `catch` which is used when there are errors in ajax.

Ahmedgagan: https://user-images.githubusercontent.com/35112518/106352996-3fbce200-630d-11eb-91a5-e30cf77cc0bc.mov

The issue we were facing in the `then` method. when we drop the reaction from the button, it **re-appeared** due to the condition `if(value === undefined)` was always true, even when the backend doesn't return any error. to solve this I used `catch` which will be called only when there's an error in the backend.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
jjaffeux: yes understood, but still looks at what I do with later/cancel please

Ahmedgagan: Used cancel in the latest commit😅 do have a look

jjaffeux (approved):

PR #11, opened 2021-02-09
FEATURE: New badge added for first reaction

Ahmedgagan:
@jjaffeux


jjaffeux:
can you think of a way to write a spec for this behavior ?

Ahmedgagan:
The for query OR the whole fixture😅?

jjaffeux:
The behavior you are trying to create with this.

- getting the badge
- backfilling the badges
- ...

jjaffeux (approved):

PR #21, opened 2021-02-24
UX: Fixed Reaction animation for mobile

Ahmedgagan:


Ahmedgagan:
@jjaffeux

jjaffeux:
sorry I don't get from the title what it fixes

Ahmedgagan:
> sorry I don't get from the title what it fixes

Reaction animation in mobile was slightly off, so added a fix.
Also, the above commits were for the fix you did yesterday and I was not knowing, so removed all other code and just kept fix for animation in mobile

jjaffeux:
> > sorry I don't get from the title what it fixes
> 
> Reaction animation in mobile was slightly off, so added a fix.
> Also, the above commits were for the fix you did yesterday and I was not knowing, so removed all other code and just kept fix for animation in mobile

I mean you should update your commit  title

PR #22, opened 2021-02-25
UI: Changed UI of reactions-summary dialog box

Ahmedgagan:
@jjaffeux

### DIFF
--- a/assets/stylesheets/common/discourse-reactions.scss
+++ b/assets/stylesheets/common/discourse-reactions.scss
@@ -130,60 +130,98 @@ html.discourse-reactions-no-select {
   }
 
   .counters {
-    border-top: 1px solid var(--primary-low);
     display: grid;
-    grid-template-columns: repeat(6, 1fr);
+    grid-template-columns: repeat(1, 1fr);
+    grid-column-gap: 0.5em;
+    grid-row-gap: 0.5em;
+    padding: 0.5em;
+    margin-bottom: auto;
+  }
+}
+
+.discourse-reactions-state-panel-reaction-users {
+  .user-container {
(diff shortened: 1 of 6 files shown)


assets/stylesheets/common/discourse-reactions.scss
```diff
@@ -130,60 +130,103 @@ html.discourse-reactions-no-select {
...
+    justify-content: space-between;
+  }
+
+  .count {
+    color: var(--primary);
+    font-size: $font-up-1;
+    margin-left: 0.5em;
+  }
+
+  .hide-users {
```
jjaffeux, line 174: why all the !important here ?

assets/stylesheets/common/discourse-reactions.scss
```diff
@@ -130,60 +130,103 @@ html.discourse-reactions-no-select {
...
-      color: var(--primary-high);
-    }
+  .reaction-wrapper {
+    display: flex;
+    grid-column: 1 / 4;
   }
 
-  &:not(.is-displayed) {
-    filter: grayscale(1);
+  .show-users {
```
jjaffeux, line 214: why all the !important here ?

assets/stylesheets/common/discourse-reactions.scss
Ahmedgagan: Because It was using its specific CSS colors and was not accessing our CSS

jjaffeux: Yes I know what's the effect of important is :) I will have look after merging and see if I can't change this, maybe specificity in core is too high.

PR #24, opened 2021-02-26
DEV: Limit reaction_users to 27

Ahmedgagan:
@jjaffeux Did 27 here because
- when I drop a reaction, in the reaction-state-panel if there are users more than say 53, it should display `26 and 27 others...` but the issue here is I dropped a reaction which means 1 user is removed from 26 and it displays 25 users instead of 26.

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
@@ -317,6 +317,11 @@ export default createWidget("discourse-reactions-actions", {
           return false;
         } else if (reaction.id === post.current_user_reaction.id) {
           post.reactions[index].count -= 1;
+          const users = post.reactions[index].users;
+          users.splice(
+            users.indexOf(users.findBy("username", "ahmedgagan6")),
+            1
+          );
           return false;
         }
 
@@ -334,7 +339,7 @@ export default createWidget("discourse-reactions-actions", {
       post.reactions.every((reaction, index) => {
         if (reaction.id === attrs.reaction) {
           post.reactions[index].count += 1;
-          post.reactions[index].users.push({
+          post.reactions[index].users.unshift({
             username: this.currentUser.username,
             avatar_template: this.currentUser.avatar_template,
             can_undo: true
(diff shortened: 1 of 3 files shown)


assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -317,6 +317,11 @@ export default createWidget("discourse-reactions-actions", {
           return false;
         } else if (reaction.id === post.current_user_reaction.id) {
           post.reactions[index].count -= 1;
+          const users = post.reactions[index].users;
+          users.splice(
+            users.indexOf(users.findBy("username", "ahmedgagan6")),
```
jjaffeux: ? 😅

assets/javascripts/discourse/widgets/discourse-reactions-state-panel-reaction.js.es6
```diff
@@ -17,6 +17,8 @@ export default createWidget("discourse-reactions-state-panel-reaction", {
 
   html(attrs) {
     const elements = [];
+    const MAX_USERS_COUNT = 26;
```
jjaffeux: make it top level

plugin.rb
```diff
@@ -62,12 +62,13 @@ class Engine < ::Rails::Engine
   end
 
   add_to_serializer(:post, :reactions) do
+    MAX_USERS_COUNT = 26
```
jjaffeux: make it top level

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
Ahmedgagan: when a user drops a reaction, we were not removing the reaction-user, so removed it here

jjaffeux: read the code again please :p

jjaffeux: You hardcoded your username: ahmedgagan6

Ahmedgagan: oooh got it😂  a real mess😜

jjaffeux (approved):

PR #25, opened 2021-02-26
FIX: Improved reaction animation from button.

Ahmedgagan:
@jjaffeux


Ahmedgagan:
here the issue was, we were changing the icon when animating down, I just refactored it as the core, which doesn't change the icon to get a seamless experience

PR #28, opened 2021-03-03
DEV: Removed unwanted css from discourse-reactions-list-emoji file

Ahmedgagan:
@jjaffeux


jjaffeux:
much better thanks, I still think there's probably more everywhere, hunt for this as much as you can

PR #30, opened 2021-03-03
FEATURE: Added lazy loading to load reaction-users only when needed.

Ahmedgagan:
@jjaffeux

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -52,32 +52,24 @@ module DiscourseReactions
 
       reaction_users = {}
 
-      if reaction_value && reaction_value == DiscourseReactions::Reaction.main_reaction_id
-        likes = post.post_actions.where("deleted_at IS NULL AND post_action_type_id = ?", PostActionType.types[:like])
+      likes = post.post_actions.where("deleted_at IS NULL AND post_action_type_id = ?", PostActionType.types[:like]) if !reaction_value || reaction_value == DiscourseReactions::Reaction.main_reaction_id
 
-        if !likes.blank?
-          reaction_users[DiscourseReactions::Reaction.main_reaction_id] = {
-            users: likes.includes([:user]).limit(MAX_USERS_COUNT + 1).map { |like| { username: like.user.username, name: like.user.name, avatar_template: like.user.avatar_template, can_undo: guardian.can_delete_post_action?(like) } },
-          }
-        end
-      elsif reaction_value
-        post.reactions.where(reaction_value: reaction_value).select { |reaction| reaction[:reaction_users_count] }.each do |reaction|
-          reaction_users[reaction.reaction_value] = {
-            users: reaction.reaction_users.includes(:user).order("discourse_reactions_reaction_users.created_at desc").limit(MAX_USERS_COUNT + 1).map { |reaction_user| { username: reaction_user.user.username, name: reaction_user.user.name, avatar_template: reaction_user.user.avatar_template, can_undo: reaction_user.can_undo? } }
-          }
-        end
-      else
+      like_users = {
+        users: likes.includes([:user]).limit(MAX_USERS_COUNT + 1).map { |like| { username: like.user.username, name: like.user.name, avatar_template: like.user.avatar_template, can_undo: guardian.can_delete_post_action?(like) } },
+      } if !likes.blank?
+
+      reaction_users[DiscourseReactions::Reaction.main_reaction_id] = like_users if like_users
+
+      if !reaction_value
         post.reactions.select { |reaction| reaction[:reaction_users_count] }.each do |reaction|
           reaction_users[reaction.reaction_value] = {
             users: reaction.reaction_users.includes(:user).order("discourse_reactions_reaction_users.created_at desc").limit(MAX_USERS_COUNT + 1).map { |reaction_user| { username: reaction_user.user.username, name: reaction_user.user.name, avatar_template: reaction_user.user.avatar_template, can_undo: reaction_user.can_undo? } }
           }
         end
-
-        likes = post.post_actions.where("deleted_at IS NULL AND post_action_type_id = ?", PostActionType.types[:like])
-
-        if !likes.blank?
-          reaction_users[DiscourseReactions::Reaction.main_reaction_id] = {
-            users: likes.includes([:user]).limit(MAX_USERS_COUNT + 1).map { |like| { username: like.user.username, name: like.user.name, avatar_template: like.user.avatar_template, can_undo: guardian.can_delete_post_action?(like) } },
+      elsif reaction_value && reaction_value != DiscourseReactions::Reaction.main_reaction_id
+        post.reactions.where(reaction_value: reaction_value).select { |reaction| reaction[:reaction_users_count] }.each do |reaction|
+          reaction_users[reaction.reaction_value] = {
+            users: reaction.reaction_users.includes(:user).order("discourse_reactions_reaction_users.created_at desc").limit(MAX_USERS_COUNT + 1).map { |reaction_user| { username: reaction_user.user.username, name: reaction_user.user.name, avatar_template: reaction_user.user.avatar_template, can_undo: reaction_user.can_undo? } }
           }
         end
       end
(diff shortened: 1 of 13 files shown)


Ahmedgagan:
Didn't remove data from the serializer in this PR because if we remove it the plugin will stop working.
The next PR will be of front-end handling of this. In that, I will remove data from the serializer.

jjaffeux:
We need specs for this

app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -39,6 +42,49 @@ def reactions_received
...
+      post_id = params.require(:post_id).to_i
+      reaction_value = params[:reaction_value]
+
+      post = Post.find_by(id: post_id)
+
+      raise Discourse::InvalidParameters if !post || (reaction_value && !DiscourseReactions::Reaction.valid_reactions.include?(reaction_value))
+
+      reaction_users = {}
+
+      if reaction_value && reaction_value == DiscourseReactions::Reaction.main_reaction_id
```
jjaffeux: I see a lot of duplication in all these blocks, try to DRY if possible

Ahmedgagan:
@jjaffeux added a new commit with improved code quality. does it look nice now?

app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -39,6 +42,41 @@ def reactions_received
...
+      like_users = {
+        users: likes.includes([:user]).limit(MAX_USERS_COUNT + 1).map { |like| { username: like.user.username, name: like.user.name, avatar_template: like.user.avatar_template, can_undo: guardian.can_delete_post_action?(like) } },
+      } if !likes.blank?
+
+      reaction_users[DiscourseReactions::Reaction.main_reaction_id] = like_users if like_users
+
+      if !reaction_value
+        post.reactions.select { |reaction| reaction[:reaction_users_count] }.each do |reaction|
+          reaction_users[reaction.reaction_value] = {
+            users: reaction.reaction_users.includes(:user).order("discourse_reactions_reaction_users.created_at desc").limit(MAX_USERS_COUNT + 1).map { |reaction_user| { username: reaction_user.user.username, name: reaction_user.user.name, avatar_template: reaction_user.user.avatar_template, can_undo: reaction_user.can_undo? } }
```
jjaffeux: This block is exactly the same than the other one and has lots of logic, share it

plugin.rb
```diff
@@ -61,6 +61,7 @@ class Engine < ::Rails::Engine
     put "/discourse-reactions/posts/:post_id/custom-reactions/:reaction/toggle" => "custom_reactions#toggle", constraints: { format: :json }
     get "/discourse-reactions/posts/my-reactions" => "custom_reactions#my_reactions", as: "my_reactions"
     get "/discourse-reactions/posts/reactions-received" => "custom_reactions#reactions_received", as: "reactions_received"
+    get "/discourse-reactions/:post_id/reactions-users" => "custom_reactions#post_reactions_users", as: "post_reactions_users"
```
jjaffeux: Why don't we just add this on posts route ?

`/posts/:id/reactions`

Also I'm unclear why we call it "reactions users", it's reactions no ?

Ahmedgagan: Sure that's cool, ill add it to posts route
We are returning users of reactions here. so I think reactions-users sounds good. thoughts?

jjaffeux: I don't know I feel like we are just return reactions, and reactions have users and other stuff.... so reactions seems good enough to me no ?

Ahmedgagan: Yes, that's right we are returning reactions with users. But the reason we are creating this route was for lazy loading of reaction-users and that's where this route will be used. so naming it as per its use was what I thought. Thoughts?

Ahmedgagan:
@jjaffeux Added all the changes as per comments here. Also improved its RSpec.

Ahmedgagan:
@jjaffeux may i add the frontend code in the same PR?

jjaffeux:
Sure, will be easier to test

Ahmedgagan:
@jjaffeux Implemented the lazy loading. Now, we need to remove reactions from the serializer.
Do try it out, if you get some time And I'll start removing users from post-serializer.

Ahmedgagan:
@jjaffeux Removed users from post-serializer in the above commits

assets/javascripts/discourse/widgets/discourse-reactions-state-panel.js.es6
```diff
@@ -55,20 +55,23 @@ export default createWidget("discourse-reactions-state-panel", {
...
+      "div.counters",
+      attrs.post.reactions.map(reaction =>
+        this.attach("discourse-reactions-state-panel-reaction", {
+          reaction,
+          users: attrs.state[reaction.id],
+          post: attrs.post,
+          isDisplayed: reaction.id === this.state.displayedReactionId
+        })
+      )
+    ) : h('div.center', h('div.spinner'));
```
jjaffeux: I don't like this center, it describes the expected behavior, not what this object is about.

html should be decorelated from css, what if you do not want to have it centered in another theme ?

assets/javascripts/discourse/widgets/discourse-reactions-state-panel.js.es6
```diff
@@ -55,20 +55,23 @@ export default createWidget("discourse-reactions-state-panel", {
...
-          "div.counters",
-          attrs.post.reactions.map(reaction =>
-            this.attach("discourse-reactions-state-panel-reaction", {
-              reaction,
-              post: attrs.post,
-              isDisplayed: reaction.id === this.state.displayedReactionId
-            })
-          )
-        )
+        divContainer
```
jjaffeux: I would also prefer a better name than divContainer, I don't really care about the html implementation here, what is this container about? what does it contain?

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
```diff
@@ -10,14 +12,77 @@ export default createWidget("discourse-reactions-counter", {
...
+    state[this.siteSettings.discourse_reactions_reaction_for_like] = [];
+    state.statePanelExpanded = false;
+    state.postId = null;
+    state.reactionValue = null;
+    state.postIds = [];
+
+    return state;
+  },
+
+  getUsers(reactionValue = null) {
```
jjaffeux: do we need the`= null` here ? what does it give us compared to just `getUsers(reactionValue)` ?

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
```diff
@@ -74,7 +139,8 @@ export default createWidget("discourse-reactions-counter", {
         this.attach(
           "discourse-reactions-state-panel",
           Object.assign({}, attrs, {
-            statePanelExpanded: this.state.statePanelExpanded
+            statePanelExpanded: this.state.statePanelExpanded,
+            state: this.state
```
jjaffeux, line 147: not really liking this, sharing state like sounds like a smell and might lead to zombie objects... We can keep it for now as I understand the hard problem it solves. But we need something better.

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
Ahmedgagan: When we don't use `null` here and call the function without parameter, let's say `getUsers()` then `reactionValue` is undefined and we always need to check a condition `reactionValue === undefined` rather than `!reactionValue`

jjaffeux: why ?

Capture d’écran 2021-03-08 à 15 01 15

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
Ahmedgagan: I know it's weird but I didn't found any better way to do this😅

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
Ahmedgagan: My bad, I got some misconceptions

assets/javascripts/discourse/widgets/discourse-reactions-state-panel.js.es6
Ahmedgagan: that's true will change it

Ahmedgagan:
@jjaffeux added a commit with suggested changes here

PR #31, opened 2021-03-04
REFACTOR: Sorting of reactions done from backend

Ahmedgagan:
@jjaffeux

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-state-panel.js.es6
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-state-panel.js.es6
@@ -43,15 +43,13 @@ export default createWidget("discourse-reactions-state-panel", {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


jjaffeux:
I need a spec for this, also I think it should just be a defaut scope of reactions

Ahmedgagan:
> I need a spec for this, also I think it should just be a defaut scope of reactions

I think by making it default scope it won't make a difference, as we are using likes as a separate action. We will always need to sort the array of combined(like+reactions) again.

Also, I'll add the spec for this

jjaffeux:
> > I need a spec for this, also I think it should just be a defaut scope of reactions
> 
> I think by making it default scope it won't make a difference, as we are using likes as a separate action. We will always need to sort the array of combined(like+reactions) again.
> 
> Also, I'll add the spec for this

Yes indeed... hmm... ok let's do it like this for now

jjaffeux (approved):

PR #37, opened 2021-03-10
FIX: Ignored RecordNotUnique error

Ahmedgagan:
@jjaffeux

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -12,7 +12,12 @@ module DiscourseReactions
       end
 
       publish_change_to_clients!(post)
-      DiscourseReactions::ReactionManager.new(reaction_value: params[:reaction], user: current_user, guardian: guardian, post: post).toggle!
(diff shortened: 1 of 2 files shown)


Ahmedgagan:
@jjaffeux can you suggest to me how to write the spec for this? 
As it's not defined when will the backend raise this issue, I removed this recuse block and fired 20 requests together in a spec file but all requests were successful.
Also, I checked out in the core there's no spec for this type of ignoring rescue block.

Screenshot 2021-03-10 at 12 54 09 PM

jjaffeux:
why is it closed ?

Ahmedgagan:
> why is it closed ?

Sorry for the mess, there was some issue in managing the branch. I have added a new PR for this here
https://github.com/discourse/discourse-reactions/pull/38

jjaffeux:
The way you write a spec for this:
- first you remove the code you added, this way you can ensure there was an issue before
- in  your spec I would image you attempt to create this record and yen call toggle, it should raise an error no?
- you can use `expect { # code }.to_not raise_error`

PR #38, opened 2021-03-10
FIX: Handled ReacordNotUnique error

Ahmedgagan:
Sorry for the multiple PR. It was due to some issue, now it's fixed.
@jjaffeux can you suggest to me how to write the spec for this? As it's not defined when will the backend raise this issue, I removed this recuse block and fired 20 requests together in a spec file but all requests were successful. Also, I checked out in the core there's no spec for this type of ignoring rescue block.

Screenshot 2021-03-10 at 12 54 09 PM


jjaffeux:
:+1: check my comment on previous PR

jjaffeux (approved):

PR #39, opened 2021-03-10
DEV: Bind rake task to run only on development environment

Ahmedgagan:
@jjaffeux

### DIFF
--- a/lib/tasks/reactions.rake
+++ b/lib/tasks/reactions.rake
@@ -22,7 +22,10 @@ end
 
 desc "create users and generate random reactions on a post"
 task "reactions:generate", [:post_id, :reactions_count, :reaction] => [:environment] do |_, args|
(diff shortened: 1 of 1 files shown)


jjaffeux (changes requested):
hmmm that seems weird, feel like you are assigning value here and not checking value

Ahmedgagan:
> hmmm that seems weird, feel like you are assigning value here and not checking value

That's true. added a fix for this

jjaffeux:
Sorry I should have said it before but we should display some basic error message too.

Ahmedgagan:
I have added the message in `raise` 😅 do we need any other message too?

Ahmedgagan:
Screenshot 2021-03-11 at 1 59 10 AM

jjaffeux (approved):

PR #41, opened 2021-03-10
DEV: Improved loading of reaction-users

Ahmedgagan:
@jjaffeux  Improved loading of reaction-users when reaction is not in site-setting `discourse_reactions_enabled_reactions`

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -70,9 +70,7 @@ module DiscourseReactions
       reaction_value = params[:reaction_value]
       post = Post.find_by(id: id)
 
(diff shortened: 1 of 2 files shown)


jjaffeux:
What is the goal of this change exactly?

Ahmedgagan:
Last night i was playing with the plugin, what i did was,set the siteSetting `discourse_reactions_enabled_reactions` to `open_mouth|thumbsup`.

Then added thumbsup reactions to a  post.

Then set `discourse_reactions_enabled_reactions` setting to `open_mouth|thumbsdown`

The i visited the post.
Everything was okay but reaction_users of `thumbsup` were not loading due to they were not in the state of widget also backend only allowed reactions from `discourse_reactions_enabled_reactions`

So to fix this i added this PR

PR #42, opened 2021-03-12
UX: implements the desaturated site setting

jjaffeux:


jjaffeux:
@Ahmedgagan what you think ?

Ahmedgagan:
Don't you think this increases the visual bulk? normally our eyes are set to see colorful emoji's, and by converting them to grey somehow I think it catches my eye over there.

Also, if we want this I think we can grey the reaction-emoji too. as it seems weird to have 2 color emojis in one panel. 

 thoughts?

Screenshot 2021-03-12 at 5 49 54 PM

jjaffeux:
Thats something people have been requesting. 🤷‍♂️



Envoyé de mon iPhone

Ahmedgagan:
Can we also desaturate reaction button emoji?

jjaffeux:
Im unsure about this, it would be too different from classic likes.

Envoyé de mon iPhone

Ahmedgagan:
Thats true, also, I just found we are keeping the setting for desaturated false by default, this seems to good idea for this, if user wants saturated they can trun on this setting.

I think we are good to go for this😅

PR #46, opened 2021-03-17
DEV: group publish changes

jjaffeux:


jjaffeux:
@Ahmedgagan I thought it was weird we were doing a publish change before doing anything, don't you think ?

Ahmedgagan:
I think there's a problem in our logic.  When we toggle, one reaction is dropped and other is created. May i handle this case in a new PR?

jjaffeux:
sure go ahead

PR #47, opened 2021-03-17
REFACTOR: s/is_reacted_by_user/reaction_user

jjaffeux:


jjaffeux:
I don't like the previous name `is_reacted_by_user`, any objection @Ahmedgagan ?

Ahmedgagan:
> I don't like the previous name `is_reacted_by_user`, any objection @Ahmedgagan ?

No, `reaction_user` is a perfect name😅

PR #49, opened 2021-03-17
FIX: Message bus will load both dropped and new reacted reaction

Ahmedgagan:
@jjaffeux This will load reaction_users of both dropped and newly added reaction. Also, reason behind keeping the message bus before reaction is, we want both reaction values, the one which will be dropped and the one which will be added. so if we keep it after the `toggle` we won't be able to get which reaction was dropped here. thoughts?

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -115,10 +115,19 @@ module DiscourseReactions
     end
 
     def publish_change_to_clients!(post)
+      reactions = [params[:reaction]]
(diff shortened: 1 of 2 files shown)


jjaffeux:
Imt not much linking that we might do so many request when someone reacts to a post.

Imagine 1000 users watching a post 100 liking it in few seconds.... how many requests will it do??!!

Can’t have a more optimized request to just update counters?

Ahmedgagan:
For the requests, we are making a request only when the user has hovered on a specific emoji. I think the number of requests won't be that much. thoughts?

Also, the reason for doing this is when the user hovered on an emoji means he is watching who reacted with that emoji. so, for that user, we should update the reaction-list in the background.

PR #51, opened 2021-03-18
FEX: selector raising error when reaction name has special characters

Ahmedgagan:
@jjaffeux


jjaffeux:
Interesting solution :+1:

PR #52, opened 2021-03-18
DEV: Front-end sorting of reactions when adding a new reaction to list

Ahmedgagan:
@jjaffeux 

Here, I've added front-end sorting only when adding a new reaction to the list, because IMO all other cases just shifts position of reaction which looks nice and this case seems weird because first we add a reaction to the list, and then sorting is done.

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
@@ -369,17 +369,19 @@ export default createWidget("discourse-reactions-actions", {
           count: 1
         };
 
-        const tempReactions = Object.assign([], post.reactions,);
+        const tempReactions = Object.assign([], post.reactions);
 
         tempReactions.push(newReaction);
 
-        const newReactionIndex = tempReactions.sort((reaction1, reaction2) =>
-          reaction2.count - reaction1.count || reaction1.id > reaction2.id
-            ? 1
-            : reaction2.id > reaction1.id
-            ? -1
-            : 0
-        ).indexOf(newReaction);
+        const newReactionIndex = tempReactions
+          .sort((reaction1, reaction2) =>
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -359,11 +363,27 @@ export default createWidget("discourse-reactions-actions", {
...
           count: 1
-        });
+        };
+
+        const tempReactions = Object.assign([], post.reactions);
+
+        tempReactions.push(newReaction);
+
+        const newReactionIndex = tempReactions
+          .sort((reaction1, reaction2) =>
```
jjaffeux: all this code is not readable at all

Ahmedgagan: I know but we want to sort array by 2 variables. One is count and another is string. In order to do that this was the best solution i found😅 can you suggest any improvements?

assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -373,14 +373,16 @@ export default createWidget("discourse-reactions-actions", {
...
         const newReactionIndex = tempReactions
-          .sort((reaction1, reaction2) =>
-            reaction2.count - reaction1.count || reaction1.id > reaction2.id
-              ? 1
-              : reaction2.id > reaction1.id
-              ? -1
-              : 0
-          )
+          .sort((reaction1, reaction2) => {
+            if (reaction1.count > reaction2.count) return -1;
```
Ahmedgagan, line 379: @jjaffeux thoughts on the readability of this code?

jjaffeux: yes much better

PR #53, opened 2021-03-19
DEV: adds a rake task to convert reactions to like

Ahmedgagan:
@jjaffeux 
```
usage:
rake "reactions:nuke[heart|angry]" OR
rake "reactions:nuke"
```


lib/tasks/reactions.rake
```diff
@@ -50,3 +50,32 @@ task "reactions:generate", [:post_id, :reactions_count, :reaction] => [:environm
...
+  end
+
+  reactions.each do |reaction|
+    puts "Converting '#{reaction.reaction_value}' of post_id: #{reaction.post_id} to like..."
+
+    reaction.reaction_users.each do |reaction_user|
+      post = Post.find_by(id: reaction_user.post_id)
+      user = User.find_by(id: reaction_user.user_id)
+
+      next unless post && user
```
jjaffeux, line 100: we should add some logging here:

"Couldn’t find user with id: xx, continuing to next reaction..." and same for post

Ahmedgagan:
@jjaffeux Added commits with revoking of badges and deleting data from `reaction` & `reaction_users` table with many barriers

lib/tasks/reactions.rake
```diff
@@ -50,3 +50,54 @@ task "reactions:generate", [:post_id, :reactions_count, :reaction] => [:environm
       .toggle!
   end
 end
+
+desc "Converts reactions to like"
+task "reactions:nuke", [:reaction_list_to_convert] => [:environment] do |_, args|
+  puts "Disabling the discourse_reactions plugin"
```
jjaffeux, line 63: we need a prompt confirmation here

Ahmedgagan: May I add a variable like `:disable_plugin` if the value is provided then we will consider the value, and if not provided disable the plugin?

jjaffeux: no, what I mean was we need a confirmation at the start of the task.

"You are about to destroy all reactions from database, are you sure ? y/n"

Ahmedgagan: Okay I got it😅

spec/tasks/reactions_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  fab!(:reaction_2) { Fabricate(:reaction, post: post_2, reaction_value: "open_mouth") }
+  fab!(:reaction_3) { Fabricate(:reaction, post: post_2, reaction_value: "hugs") }
+  fab!(:reaction_user_1) { Fabricate(:reaction_user, reaction: reaction_1, user: user_2, post: post_2) }
+  fab!(:reaction_user_2) { Fabricate(:reaction_user, reaction: reaction_1, user: user_1, post: post_2) }
+  fab!(:reaction_user_3) { Fabricate(:reaction_user, reaction: reaction_3, user: user_4, post: post_2) }
+  fab!(:reaction_user_4) { Fabricate(:reaction_user, reaction: reaction_2, user: user_3, post: post_2) }
+
+  before do
+    Rake::Task.clear
+    Discourse::Application.load_tasks
```
Ahmedgagan: @jjaffeux I am stuck here, `load_tasks` doesn't load the plugin's rake tasks. I am not getting a way to load it.

It gives an error
```
RuntimeError:
       Don't know how to build task 'reactions:nuke'
```
This is because it doesn't loads the plugin's rake task.

jjaffeux: try to manually `load` the task file at the top of your file

Ahmedgagan: I loaded the rake file in the plugin.rb & it worked 😅

lib/tasks/reactions.rake
```diff
@@ -50,3 +50,75 @@ task "reactions:generate", [:post_id, :reactions_count, :reaction] => [:environm
...
+  SiteSetting.post_undo_action_window_mins = POST_UNDO_ACTION_WINDOW_MINS
+
+  badge = Badge.find_by(name: I18n.t("badges.first_reaction.name"))
+
+  if badge
+    puts "Revoking '#{I18n.t("badges.first_reaction.name")}' badge from all users..."
+    BadgeGranter.revoke_all(badge)
+  end
+
+  puts "Deleting all remaining reactions and reaction_users..."
```
Ahmedgagan, line 119: @jjaffeux here can we ask admin whether he wants to delete data or not?
If he doesn't allow to delete then also nothing will break and risk is also reduced.

PR #54, opened 2021-03-21
UX: make defaults be positive and encourage constructive criticism

robbyoconnor:
Let's promote more positivity and less negative reactions by default. These can be overridden but I can't see this as a bad idea to have.


SamSaffron:
I get the reasoning here, but lets keep the discussion on meta regarding defaults. I am comfortable with where it is at now.

robbyoconnor:
@SamSaffron I figured no harm in making the PR

PR #61, opened 2021-04-02
UX: Adds visibility to state-panel when expanded

Ahmedgagan:
@jjaffeux


jjaffeux:
Why do we need this already please?

Ahmedgagan:
> Why do we need this already please?

When we use the post-badges theme-component state-panel is not displayed. Last time we did for reaction-picker but forgot to add for state-panel.

PR #67, opened 2021-04-08
DEV: Prettier, eslint, ember-template-lint

Ahmedgagan:
@jjaffeux


Ahmedgagan:
@jjaffeux some of the backend tests are failing will be adding another PR for that

PR #70, opened 2021-04-09
DEV: Fixes enum & post_serializer_spec test

Ahmedgagan:
@jjaffeux Now the validation won't fail, `DiscourseReactions::Reaction.valid_reactions` always haves the value of `discourse_reactions_reaction_for_like` setting in it.

### DIFF
--- a/lib/reaction_for_like_site_setting_enum.rb
+++ b/lib/reaction_for_like_site_setting_enum.rb
@@ -10,8 +10,8 @@ class ReactionForLikeSiteSettingEnum < EnumSiteSetting
   end
 
   def self.values
-    @values ||= begin
-      reactions = SiteSetting.discourse_reactions_enabled_reactions.split('|').map do |reaction|
+    @values = begin
+      reactions = DiscourseReactions::Reaction.valid_reactions.map do |reaction|
         { name: reaction, value: reaction }
       end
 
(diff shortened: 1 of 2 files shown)


lib/reaction_for_like_site_setting_enum.rb
```diff
@@ -3,19 +3,15 @@
...
-  HEART ||= 'heart'
-
   def self.valid_value?(val)
     values.any? { |v| v[:value] == val }
   end
 
   def self.values
-    @values ||= begin
-      reactions = SiteSetting.discourse_reactions_enabled_reactions.split('|').map do |reaction|
+    @values = begin
```
Ahmedgagan, line 11: removed `||` because values of valid_reactions can change and that causes the issue.

lib/reaction_for_like_site_setting_enum.rb
```diff
@@ -3,19 +3,15 @@
...
+      reactions = DiscourseReactions::Reaction.valid_reactions.map do |reaction|
         { name: reaction, value: reaction }
       end
-
-      [{ name: HEART, value: HEART }].concat(reactions)
```
Ahmedgagan, line 18: Heart value is already present in `DiscourseReactions::Reaction.valid_reactions`

PR #72, opened 2021-04-11
DEV: combines like_users & reaction_users both if reaction is main

Ahmedgagan:
@jjaffeux combines like_users & reaction_users both if reaction is main in `post_reactions_users` / `reactions-users` route

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -81,12 +81,11 @@ module DiscourseReactions
         count = likes.length
         users = format_likes_users(likes)
 
-        if main_reaction[:reaction_users_count]
+        if main_reaction && main_reaction[:reaction_users_count]
           (users << get_users(main_reaction)).flatten!
+          users.sort_by! { |user| user[:created_at] }
           count += main_reaction.reaction_users_count.to_i
         end
-
-        users.sort_by! { |user| user[:created_at] }
 
         reaction_users << {
           id: DiscourseReactions::Reaction.main_reaction_id,
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -78,7 +78,7 @@ module DiscourseReactions
 
       if likes.present?
         main_reaction = DiscourseReactions::Reaction.find_by(reaction_value: DiscourseReactions::Reaction.main_reaction_id, post_id: post.id)
-        count = likes.length.to_i
+        count = likes.length
         users = format_likes_users(likes)
 
         if main_reaction[:reaction_users_count]
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -77,15 +77,26 @@ module DiscourseReactions
       likes = post.post_actions.where('deleted_at IS NULL AND post_action_type_id = ?', PostActionType.types[:like]) if !reaction_value || reaction_value == DiscourseReactions::Reaction.main_reaction_id
 
       if likes.present?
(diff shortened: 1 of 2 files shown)


jjaffeux:
Please write a spec exhibiting what behavior is this changing/fixing

Ahmedgagan:
> Please write a spec exhibiting what behavior is this changing/fixing

Sure, I'll write a spec for this behavior

Ahmedgagan:
@jjaffeux Added the rspec for this behavior

jjaffeux:
Is my fix in the serializer still need with this?

Ahmedgagan:
Yes both are required, as we call this route only when anyone hover's on panel OR open the panel

app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -77,15 +77,26 @@ def post_reactions_users
       likes = post.post_actions.where('deleted_at IS NULL AND post_action_type_id = ?', PostActionType.types[:like]) if !reaction_value || reaction_value == DiscourseReactions::Reaction.main_reaction_id
 
       if likes.present?
+        main_reaction = DiscourseReactions::Reaction.find_by(reaction_value: DiscourseReactions::Reaction.main_reaction_id, post_id: post.id)
+        count = likes.length.to_i
```
jjaffeux: why do we need to_i  here ? can length be something else than integer ?

Ahmedgagan: No😅 nice catch Ill change it

spec/requests/custom_reactions_controller_spec.rb
```diff
@@ -155,6 +155,21 @@
     end
   end
 
+  context 'changing discourse_reactions_like_icon' do
+    before do
+      SiteSetting.discourse_reactions_reaction_for_like = 'laughing'
+    end
+
+    it 'merges identic custom reaction into likes' do
+      user_5 = Fabricate(:user)
```
jjaffeux: why user 5 ? can you use let for this ?

spec/requests/custom_reactions_controller_spec.rb
```diff
@@ -155,6 +155,21 @@
...
+      like = Fabricate(:post_action, post: post_2, user: user_5, post_action_type_id: PostActionType.types[:like])
```
jjaffeux: same here

spec/requests/custom_reactions_controller_spec.rb
```diff
@@ -155,6 +155,21 @@
     end
   end
 
+  context 'changing discourse_reactions_like_icon' do
```
jjaffeux: are we changing the icon or the reaction? Be very careful with wording please

Ahmedgagan: oooh my bad we are changing `reaction`

Ahmedgagan:
@jjaffeux used `fab!` instead of `user_5=` & changed the context name in the latest commit

spec/requests/custom_reactions_controller_spec.rb
```diff
@@ -155,6 +155,21 @@
     end
   end
 
+  context 'changing discourse_reactions_reaction_for_like' do
```
jjaffeux: this test still doesn’t make it clear what's happening, do a before/after, eg:

initial situation assert X likes Y laughing

change setting

assert X likes and Y laughing

spec/requests/custom_reactions_controller_spec.rb
```diff
@@ -153,6 +155,24 @@
       get "/discourse-reactions/posts/1000000/reactions-users.json?reaction_value=test"
       expect(response.status).to eq(400)
     end
+
+    it 'merges identic custom reaction into likes' do
```
Ahmedgagan, line 159: @jjaffeux How does it look now?

PR #73, opened 2021-04-12
FEATURE: Adds per day report of reactions & likes

Ahmedgagan:
@jjaffeux

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -143,4 +143,74 @@ after_initialize do
   add_model_callback(User, :before_destroy) do
     DiscourseReactions::ReactionUser.where(user_id: self.id).delete_all
   end
+
+  add_report('discourse_reactions') do |report|
+    report.modes = [:table]
+
+    report.data = []
+
+    report.labels = [
+      {
+        type: :text,
+        property: :date,
+        title: I18n.t("reports.discourse_reactions.labels.date")
+      },
+      {
+        type: :integer,
+        property: :like_count,
+        title: I18n.t("reports.discourse_reactions.labels.likes")
+      }
+    ]
+
+    titles = DiscourseReactions::Reaction.where("reaction_users_count IS NOT NULL AND reaction_value!=?", DiscourseReactions::Reaction.main_reaction_id).pluck(:reaction_value).uniq
+
+    titles.each { |title|
+      report.labels << {
+        type: :integer,
+        property: "#{title}_count",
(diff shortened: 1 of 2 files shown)


jjaffeux:
I wonder if we could find a way to override likes_report when reactions is enabled

Ahmedgagan:
I'll take a look if it's possible 😅

Ahmedgagan:
@jjaffeux Does the latest commit look okay? I tried to override the likes_report method

jjaffeux:
Looks nice 👍  now try to separate likes and custom reactions please

plugin.rb
```diff
@@ -143,4 +143,85 @@ class Engine < ::Rails::Engine
...
+            }
+          ]
+
+          titles = DiscourseReactions::Reaction.where("reaction_users_count IS NOT NULL AND reaction_value!=?", DiscourseReactions::Reaction.main_reaction_id).pluck(:reaction_value).uniq
+
+          titles.each { |title|
+            report.labels << {
+              type: :number,
+              property: "#{title}_count",
+              title: title
```
jjaffeux: Cant you just do this here:

```
PrettyText.unescape_emoji(CGI::escapeHTML(title))
```

Ahmedgagan: I tried it, But it doesn't change anything. It returns the same string even if the emoji is not available

jjaffeux:
You must be doing something wrong:

```
[1] pry(main)> PrettyText.unescape_emoji(CGI::escapeHTML(":+1:"))
=> "<img width=\"20\" height=\"20\" src='/images/emoji/twitter/+1.png?v=9' title='+1' alt='+1' class='emoji'>"
```

Like do you have : as prefix and suffix ?

Ahmedgagan:
> You must be doing something wrong:
> 
> 
> 
> ```
> 
> [1] pry(main)> PrettyText.unescape_emoji(CGI::escapeHTML(":+1:"))
> 
> => ""
> 
> ```
> 
> 
> 
> Like do you have : as prefix and suffix ?

Yes i was forgetting :

Ahmedgagan:
@jjaffeux This worked also added change to the core PR to support this change

Ahmedgagan:
@jjaffeux made it compatible with new core changes

jjaffeux:
I already did it locally :) There are multiple things not working here that I'm trying to fix.

PR #76, opened 2021-04-14
DEV: Allows '-1' types of reaction

Ahmedgagan:
@jjaffeux when there was a `-1` in the list. this regex used to remove `-` and that caused the 422 error.
Now, as we are using emoji-picker we can directly split the string from "|"


jjaffeux:
Theres a plugin test failure

Ahmedgagan:
@jjaffeux Fixed the tests

PR #78, opened 2021-04-20
UX: improves copy when editing a reaction is forbidden

jjaffeux:


config/locales/client.en.yml
```diff
@@ -13,7 +13,7 @@ en:
         too_many_request:
           one: "You have been reacting a bit too fast, try again in %{count} second."
           other: "You have been reacting a bit too fast, try again in %{count} seconds."
-        forbidden: "You acted on this post a long time ago so you are not allowed to remove your reaction now."
+        forbidden: "That reaction was done too long ago. It can no longer be modified or removed."
```
eviltrout:
```suggestion
        forbidden: "That reaction was created too long ago. It can no longer be modified or removed."
```
I feel like "created" works better than "done"

cvx (approved):

PR #80, opened 2021-04-29
FIX: move exception out of transaction to avoid poisoning

jjaffeux:
I was causing this kind of errors:

```
ActiveRecord::StatementInvalid (PG::InFailedSqlTransaction: ERROR:  current transaction is aborted, commands ignored until end of transaction block
)
```

### DIFF
--- a/app/services/discourse_reactions/reaction_manager.rb
+++ b/app/services/discourse_reactions/reaction_manager.rb
@@ -11,11 +11,11 @@ module DiscourseReactions
     end
 
     def toggle!
-      ActiveRecord::Base.transaction do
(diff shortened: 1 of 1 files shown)


jjaffeux:
@lis2 I think you wrote the original code. I don't think we need the exception and check to be in the transaction, if we do I could change it to manually rollback in a begin/rescue block around the transaction.

jjaffeux:
ping @lis2

KrisKotlarek (approved):

KrisKotlarek:
@jjaffeux I am sorry, I completely missed that github notification :(

jjaffeux:
no worries, question is still valid though 😁

PR #84, opened 2021-05-18
FIX: prevents unintentional click on scroll

jjaffeux:
A better solution would be to use touchMove but this is not implemented yet in core and would add complexity. This should be good enough for our use case.


assets/javascripts/discourse/widgets/discourse-reactions-actions.js.es6
```diff
@@ -214,7 +213,8 @@ export default createWidget("discourse-reactions-actions", {
         ) {
           this.toggleReactions(event);
         }
-      } else {
+      } else if (duration > 0 && duration < 200) {
+        // anything taking more than 200ms is most likely not intenional click
```
ZogStriP:
```suggestion
        // anything taking more than 200ms is most likely not intentional click
```

ZogStriP (approved):

PR #85, opened 2021-05-24
DEV: Adds a rake task to convert the reactions

Ahmedgagan:
@jjaffeux 
This rake task convert's the reaction to another OR drops the reaction.
This was discussed here https://meta.discourse.org/t/discourse-reactions-beyond-likes/183261/106?u=ahmed_gagan

### DIFF
--- a/lib/tasks/reactions.rake
+++ b/lib/tasks/reactions.rake
@@ -56,42 +56,45 @@ task "reactions:convert", [:old_reaction, :new_reaction] do |_, args|
   old_reaction = args[:old_reaction]
   new_reaction = args[:new_reaction]
 
-  if !old_reaction
-    raise "Please provide the reaction you want to convert i.e. 'old_reaction'"
+  begin
+    if !old_reaction
+      raise "Please provide the reaction you want to convert i.e. 'old_reaction'"
+    end
+
+    POST_UNDO_ACTION_WINDOW_MINS = SiteSetting.post_undo_action_window_mins
+
+    SiteSetting.post_undo_action_window_mins = 2000000000
+
+    if new_reaction && !DiscourseReactions::Reaction.valid_reactions.include?(new_reaction)
+      raise "Invalid Reaction, please check 'new_reaction' parameter and try again!"
+    end
+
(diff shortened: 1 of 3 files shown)


lib/tasks/reactions.rake
```diff
@@ -50,3 +50,48 @@ task "reactions:generate", [:post_id, :reactions_count, :reaction] => [:environm
...
+
+  reaction = DiscourseReactions::Reaction.find_by(reaction_value: old_reaction)
+
+  if !reaction
+    raise "Please provide an existing reaction in 'old_reaction' parameter and try again!"
+  end
+
+  puts "Converting reactions"
+  reaction.reaction_users.each do |reaction_user|
+    post = Post.find_by(id: reaction_user.post_id)
```
jjaffeux: cant you join users ands posts in reaction users and use it fo lookup post/user ?

lib/tasks/reactions.rake
```diff
@@ -50,3 +50,48 @@ task "reactions:generate", [:post_id, :reactions_count, :reaction] => [:environm
...
+
+    DiscourseReactions::ReactionManager.new(
+      reaction_value: new_reaction ? new_reaction : old_reaction,
+      user: user,
+      guardian: Guardian.new(user),
+      post: post
+    ).toggle!
+  end
+
+  SiteSetting.post_undo_action_window_mins = POST_UNDO_ACTION_WINDOW_MINS
```
jjaffeux: I would prefer if all of this was in a begin/ensure block, so we can be certain whatever happen that we are restoring the setting.

lib/tasks/reactions.rake
Ahmedgagan: I think we don't need the join, I used the associations here in the latest commit

lib/tasks/reactions.rake
Ahmedgagan: Indeed, added a commit with begin/ensure block

Ahmedgagan:
@jjaffeux i have added commits with suggested changes. Can you take a look?

PR #87, opened 2021-06-04
PERF: Fix N+1 queries when viewing a topic.

tgxworld:
This is a revert of 30e424e3b6d5c482093c4c95c4ca83c80f7eb357 which
incorrectly removed the eagar loading that has been done in the
TopicViewSerializerMixin.

Follow-up to 30e424e3b6d5c482093c4c95c4ca83c80f7eb357

### DIFF
--- a/lib/discourse_reactions/topic_view_serializer_extension.rb
+++ b/lib/discourse_reactions/topic_view_serializer_extension.rb
@@ -3,7 +3,10 @@
 module DiscourseReactions::TopicViewSerializerExtension
   def posts
     if SiteSetting.discourse_reactions_enabled
-      posts = object.posts.includes(:post_actions, reactions: { reaction_users: :user })
+      posts = object.posts
+        .includes(:post_actions, reactions: { reaction_users: :user })
+        .to_a
+
       post_ids = posts.map(&:id).uniq
 
       posts_reaction_users_count = TopicViewSerializer.posts_reaction_users_count(post_ids)
--- a/plugin.rb
+++ b/plugin.rb
@@ -75,7 +75,10 @@ after_initialize do
       }
     end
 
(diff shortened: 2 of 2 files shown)


lib/discourse_reactions/topic_view_serializer_extension.rb
```diff
@@ -3,7 +3,10 @@
 module DiscourseReactions::TopicViewSerializerExtension
   def posts
     if SiteSetting.discourse_reactions_enabled
-      posts = object.posts.includes(:post_actions, reactions: { reaction_users: :user })
+      posts = object.posts
+        .includes(:post_actions, reactions: { reaction_users: :user })
+        .to_a
```
tgxworld, line 8: We need to load this upfront and set `posts` to the loaded records. Otherwise, we end up working with an ActiveRecord association which will trigger a DB query each time we iterate through it.

plugin.rb
```diff
@@ -94,7 +97,8 @@ class Engine < ::Rails::Engine
 
   add_to_serializer(:post, :current_user_reaction) do
     return nil unless scope.user.present?
-    object.reactions.includes([:reaction_users]).each do |reaction|
```
tgxworld, line 97: This was converting an ActiveRecord collection proxy back into an association resulting in a query being triggered when the associations are being accessed.

plugin.rb
```diff
@@ -75,7 +75,10 @@ class Engine < ::Rails::Engine
       }
     end
 
-    likes = object.post_actions.where('deleted_at IS NULL AND post_action_type_id = ?', PostActionType.types[:like])
```
tgxworld, line 78: Records have already been eager loaded so we should not be calling ActiveRecord query methods on the collection proxy.

jjaffeux (approved):

PR #88, opened 2021-06-04
PERF: prevents operations on eager loaded and AR proxy

jjaffeux:


jjaffeux:
@tgxworld I have moved what seems safe in a standalone PR.

@lis2 could you have a look at the comment of @tgxworld here https://github.com/discourse/discourse-reactions/pull/87#discussion_r645402361 also after discussions with @eviltrout on chat, it seems like this is a pattern we shouldn't be using

tgxworld (approved):

PR #89, opened 2021-06-22
FIX: cleanup MessageBus subscriptions when navigating between topics

SamSaffron:
Previously we would leak subscriptions to every post you ever viewed.

Note this is just a quick fix. We also want to change it so we use a single
channel per topic.

Otherwise browsing through a 5000 post topic will result in 5000 subscriptions

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
@@ -5,6 +5,14 @@ import { createWidget } from "discourse/widgets/widget";
 import { cancel, later, schedule } from "@ember/runloop";
 import CustomReaction from "../models/discourse-reactions-custom-reaction";
 import MessageBus from "message-bus-client";
+import { addWidgetCleanCallback } from "discourse/components/mount-widget";
+
+let subscriptions = [];
+
+addWidgetCleanCallback("post-stream", () => {
+  subscriptions.forEach((channel) => MessageBus.unsubscribe(channel));
+  subscriptions = [];
+});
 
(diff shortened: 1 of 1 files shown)


arpitjalan (approved):

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
```diff
@@ -5,6 +5,14 @@ import { createWidget } from "discourse/widgets/widget";
 import { cancel, later, schedule } from "@ember/runloop";
 import CustomReaction from "../models/discourse-reactions-custom-reaction";
 import MessageBus from "message-bus-client";
+import { addWidgetCleanCallback } from "discourse/components/mount-widget";
+
+let subscriptions = [];
+
+addWidgetCleanCallback("post-stream", () => {
+  subscriptions.forEach((channel) => MessageBus.unsubscribe(channel));
```
martin-brennan, line 13: Why can't we just call `this.unsubscribe` here?

https://github.com/discourse/discourse-reactions/blob/7de5af9e4bef017e4f55cfa346d9485002fe29bf/assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6#L30-L35

martin-brennan (changes requested):
Just one question before approval

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
SamSaffron: sorry replied in chat but this is a global api so it has no access to particular instances

PR #90, opened 2021-06-23
FIX: relying on topic level event :acted is enough

jjaffeux:


ZogStriP (approved):
Wow, so we were doing all this work for "nothing"? (or at least it was duplicate work?)

jjaffeux:
So far I would say yes, there might be a case I'm not thinking of ATM, but anyways it seems basic behavior is working, so I prefer to remove it for now and see if I need to add something more specific after.

Ahmedgagan:
> So far I would say yes, there might be a case I'm not thinking of ATM, but anyways it seems basic behavior is working, so I prefer to remove it for now and see if I need to add something more specific after.

@jjaffeux Just for the reference the reason for the 2 message bus was discussed here. https://github.com/discourse/discourse-reactions/pull/49

jjaffeux:
> > So far I would say yes, there might be a case I'm not thinking of ATM, but anyways it seems basic behavior is working, so I prefer to remove it for now and see if I need to add something more specific after.
> 
> 
> 
> @jjaffeux Just for the reference the reason for the 2 message bus was discussed here. https://github.com/discourse/discourse-reactions/pull/49

Yes I saw and remember this, really doesn't strike me as an edge case worth all this complexity and perf hit.

Will see if anyone complains of this.

PR #93, opened 2021-06-25
FIX: attempts to have a topic level message bus event only

jjaffeux:


eviltrout:
Note for myself or others watching this: I had a chat with @jjaffeux about using `this.dispatch` instead of this approach.

jjaffeux:
@eviltrout I updated this with the dirty keys change

eviltrout (dismissed):

assets/javascripts/initializers/discourse-reactions.js.es6
```diff
@@ -24,6 +24,33 @@ function initializeDiscourseReactions(api) {
...
+              onRefresh: "reactionsChanged",
+              refreshArg: data,
+            }
+          );
+          this._refresh({ id: data.post_id });
+        });
+      }
+    },
+
+    unsusbcribe() {
```
tgxworld: This function doesn't seem to be called at all since core does not have an `unsubscribe` function defined.

app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -155,5 +156,22 @@ def fetch_post_from_params
...
     end
+
+    def publish_change_to_clients!(post)
+      reactions = [params[:reaction]]
+      reaction_user = DiscourseReactions::ReactionUser.find_by(user_id: current_user.id, post_id: post.id)
+
+      if reaction_user
+        reaction = DiscourseReactions::Reaction.find(reaction_user.reaction_id)
+        reactions.push(reaction.reaction_value)
+      end
```
tgxworld:
```suggestion
      reaction_id = DiscourseReactions::ReactionUser
        .where(user_id: current_user.id, post_id: post.id)
        .pluck_first(:reaction_id)

      if reaction_id
        reaction_value = DiscourseReactions::Reaction
          .where(reaction_id)
          .pluck_first(:reaction_value)
          
        reactions.push(reaction_value) if reaction_value
      end
```

We can just pluck the values instead of creating ActiveRecord objects.

assets/javascripts/initializers/discourse-reactions.js.es6
jjaffeux: yes indeed, it used to be in topic controller but we changed it to be able to use dirtyKeys.

Will see what I can do to make it called.

eviltrout (approved):

PR #97, opened 2021-07-19
FIX: updates state panel popper position after loading users

jjaffeux:


ZogStriP (approved):

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
```diff
@@ -82,6 +82,7 @@ export default createWidget("discourse-reactions-counter", {
       } else {
         this.state.postId = null;
       }
+      _popperStatePanel && _popperStatePanel.update();
```
cvx: I believe after a couple of false starts we now can use the safe navigation operator in plugins, right?

jjaffeux: Yes :) old habits :p

PR #98, opened 2021-07-19
FIX: attempts to make clicks more correct

jjaffeux:
Due to various minor issues and one major one, it could sometimes be very easy to close the panel, or very hard to open it.

The biggest issues is related to how during dev we decided to let various reactions parts be in different widgets trees, which makes global state complicated.

I have various ideas to make it better, but I think we should first attempt to move everything (visually and in code) into one widget tree. I will first try to make this accepted, before going into complex refactors.

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
@@ -89,6 +89,19 @@ export default createWidget("discourse-reactions-counter", {
 
   click(event) {
     if (!this.capabilities.touch || !this.site.mobileView) {
+      event.stopPropagation();
+
+      // in case we lost sync due to another widget not in the same tree
+      // collapsing the panel, we attempt to reconciliate from DOM state
+      const container = document.getElementById(this.buildId(this.attrs));
+      if (
+        !container
+          .querySelector(".discourse-reactions-state-panel")
+          .classList.contains("is-expanded")
+      ) {
+        this.state.statePanelExpanded = false;
+      }
+
       if (!this.state.statePanelExpanded) {
         this.getUsers();
       }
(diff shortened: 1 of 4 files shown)


assets/javascripts/discourse/widgets/discourse-reactions-state-panel-reaction.js.es6
```diff
@@ -42,15 +51,10 @@ export default createWidget("discourse-reactions-state-panel-reaction", {
...
-        this.attach("button", {
-          action: "showUsers",
-          contents: [
-            iconNode(attrs.isDisplayed ? "chevron-up" : "chevron-down"),
-          ],
-          actionParam: attrs,
-          className: "show-users",
-          title: "",
-        })
+        h(
```
jjaffeux, line 54: note we should implement bubbles=false into core widgets will make a PR about this

jjaffeux: hmmm actually I see ember deprecated bubbles... so I guess it's fine like this

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
```diff
@@ -89,6 +89,19 @@ export default createWidget("discourse-reactions-counter", {
 
   click(event) {
     if (!this.capabilities.touch || !this.site.mobileView) {
+      event.stopPropagation();
+
+      // in case we lost sync due to another widget not in the same tree
+      // collapsing the panel, we attempt to reconciliate from DOM state
+      const container = document.getElementById(this.buildId(this.attrs));
```
eviltrout, line 96: Ideally another widget would change the state and not add the class. Is this fixable?

eviltrout (approved):

assets/javascripts/discourse/widgets/discourse-reactions-counter.js.es6
jjaffeux: Yes I think it is, it's the meaning of my commit message. I will attempt to refactor this deeper.

PR #110, opened 2021-10-11
SECURITY: leaking PMs and secure categories topics

KrisKotlarek:
Ensure that user has access to private messages and categories of he topics


SamSaffron:
there has to already be something in core that does all this securing? Are you sure there is nothing you can use?

SamSaffron (approved):

ZogStriP:
I agree with @SamSaffron, we already have plenty of methods to filter out content you can't see. For example `Post.secured` or `Topic.listable_topics.visible.secured`.

PR #116, opened 2021-11-30
FEATURE: Consolidate notifications

romanrizzi:
Depends on discourse/discourse#14904


eviltrout (approved):
I'm enjoying this new API!

PR #118, opened 2021-12-09
FIX: reaction notifications were broken

arpitjalan:


SamSaffron:
Is this deliberate or a typo in the localization file?

arpitjalan:
I believe the use of plural `titles` is deliberate and added in this commit -- https://github.com/discourse/discourse-reactions/commit/1945e3b503d58ddeb81f8c809e4b4ba0bf017cdf.

martin-brennan (approved):

PR #122, opened 2021-12-21
FIX: use correct reaction notification locale key

arpitjalan:


cvx (approved):
Will this one work? 😅

arpitjalan:
@CvX It should. 🙏🏼

PR #124, opened 2021-12-23
FEATURE: Collapse multiple reactions to the same post.

romanrizzi:
When users have their like frequency set to `always`, we collapse multiple reaction notifications when users react to the same post. With this change, we reach feature parity between reactions and likes.

### DIFF
--- a/app/services/discourse_reactions/reaction_notification.rb
+++ b/app/services/discourse_reactions/reaction_notification.rb
@@ -13,16 +13,6 @@ module DiscourseReactions
 
       return if post_user.user_option.like_notification_frequency == UserOption.like_notification_frequency_type[:never]
 
-      last_notification = post_user.notifications
-        .order("notifications.id DESC")
-        .find_by(
-          topic_id: @post.topic_id,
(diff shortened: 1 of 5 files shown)


app/services/discourse_reactions/reaction_notification.rb
```diff
@@ -43,13 +33,6 @@ def delete
 
     private
 
-    def enabled_reaction_notifications?(user, notification)
-      return true if user.user_option.like_notification_frequency == UserOption.like_notification_frequency_type[:always]
-      return true if user.user_option.like_notification_frequency == UserOption.like_notification_frequency_type[:first_time_and_daily] && notification.created_at < 1.day.ago
-      return true if user.user_option.like_notification_frequency == UserOption.like_notification_frequency_type[:first_time] && notification
-      false
-    end
```
romanrizzi, line 51: We already have a PostAlerter extension for this.

romanrizzi:
Rule is similar to discourse/discourse#15393

cvx (approved):

PR #125, opened 2021-12-23
FEATURE: Remove overlap with like notifications.

romanrizzi:
Adding shadow likes also creates a "like" notification, so we now use the new PostActionCreator silent option and instead create a reaction notification. We store the icon in the notification payload.


.discourse-compatibility
```diff
@@ -1 +1,2 @@
-2.7.6: a9483ca1425c0236eccbae9260b7036795780992
\ No newline at end of file
+v2.8.0.beta10: a0508d2d5f9bd53ee59ecf9c94e4328ae6a9616b
```
gschlager:
```suggestion
2.8.0.beta10: a0508d2d5f9bd53ee59ecf9c94e4328ae6a9616b
```

pmusaraj (approved):

gschlager (changes requested):

.discourse-compatibility
romanrizzi: Good catch!

gschlager (approved):

PR #128, opened 2021-12-30
FIX: Keep the heart icon when consolidating multiple like reactions.

romanrizzi:


romanrizzi:
Depends on https://github.com/discourse/discourse/pull/15428

eviltrout (approved):

PR #134, opened 2022-01-12
FIX: Correct like reactions count

OsamaSayegh:
https://github.com/discourse/discourse-reactions/commit/5c5ab85c0d1dcec6eb5464d1ff78f6f799d44381 introduced a regression that caused double like reactions when someone likes a post. The commit has been reverted, but we need a database migration to remove the double likes that were added before the commit was reverted.

### DIFF
--- a/db/post_migrate/20220112091339_reset_erroneous_like_reactions_count.rb
+++ b/db/post_migrate/20220112091339_reset_erroneous_like_reactions_count.rb
@@ -15,13 +15,27 @@ class ResetErroneousLikeReactionsCount < ActiveRecord::Migration[6.1]
     # AR enum in the Reaction model
     emoji_reaction_type = 0
 
-    DB.exec(<<~SQL, like_reaction: like_reaction, emoji_reaction_type: emoji_reaction_type)
-      UPDATE discourse_reactions_reactions
-      SET reaction_users_count = NULL
+    inconsistent_reactions = DB.query(<<~SQL, like_reaction: like_reaction, emoji_reaction_type: emoji_reaction_type)
+      SELECT id
+      FROM discourse_reactions_reactions
       WHERE
         reaction_type = :emoji_reaction_type AND
         reaction_value = :like_reaction AND
         reaction_users_count IS NOT NULL
+    SQL
+
+    return if inconsistent_reactions.size == 0
+    ids = inconsistent_reactions.map(&:id)
+
+    DB.exec(<<~SQL, ids: ids)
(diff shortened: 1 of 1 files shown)


romanrizzi (commented):
We need to remove the reaction user rows or we'll show duplicate avatars when expanding the reaction panel

OsamaSayegh:
Ok I've updated the migration to remove `reaction_users` rows (although I couldn't repro duplicate avatars but I agree they should be removed).

db/post_migrate/20220112091339_reset_erroneous_like_reactions_count.rb
```diff
@@ -0,0 +1,45 @@
...
+    ids = inconsistent_reactions.map(&:id)
+
+    DB.exec(<<~SQL, ids: ids)
+      DELETE FROM discourse_reactions_reaction_users
+      WHERE reaction_id IN (:ids)
+    SQL
+
+    DB.exec(<<~SQL, ids: ids)
+      UPDATE discourse_reactions_reactions
+      SET reaction_users_count = NULL
```
romanrizzi, line 37: Should we set this to 0 instead? or perhaps maybe remove them?

The method that deletes unused reactions looks specifically for 0:

https://github.com/discourse/discourse-reactions/blob/main/app/services/discourse_reactions/reaction_manager.rb#L106

romanrizzi (commented):
Tested it locally and it looks good! Left one last comment.

db/post_migrate/20220112091339_reset_erroneous_like_reactions_count.rb
OsamaSayegh: I was looking at an old post in our internal instance that didn't have duplicate hearts/likes and it had `reaction_users_count = NULL`. In fact all posts that had at least a like before the regression have `reaction_users_count = NULL`:

```
[3] pry(main)> DiscourseReactions::Reaction.where("reaction_type = 0 AND reaction_value = 'heart' AND reaction_users_count IS NULL").count
=> 57253
```

I'd prefer the migration to bring all impacted posts into line with the posts that were not impacted by the regression 😄 

Edit: we have roughly 300 posts that are impacted by the regression:

```
[1] pry(main)> DiscourseReactions::Reaction.where("reaction_type = 0 AND reaction_value = 'heart' AND reaction_users_count IS NOT NULL").count
=> 314
```
If I look at `reaction_users` records for those 314 reactions, the oldest one is from yesterday when the regression was deployed.

romanrizzi (approved):

PR #141, opened 2022-02-02
FIX: Remove thumbsup reactions when there's also a +1 reaction.

romanrizzi:
The thumbsup reaction didn't work, so we recommended using +1 instead as a workaround. The post will have both reactions associated if someone reacted with a thumbsup, then it was renamed to +1, and another user reacted with a +1.

The post-migration to rename thumbsup reactions we added in a previous commit won't work in the above scenario, so we should directly remove them and leave the existing +1 reactions.

### DIFF
--- a/db/post_migrate/20220201162748_rename_thumbsup_reactions.rb
+++ b/db/post_migrate/20220201162748_rename_thumbsup_reactions.rb
@@ -19,10 +19,31 @@ class RenameThumbsupReactions < ActiveRecord::Migration[6.1]
       SQL
     end
 
-    DB.exec(<<~SQL, alias: alias_name, new_value: original_name)
+    has_both_reactions = DB.query_single(<<~SQL, alias: alias_name, new_value: original_name)
+      SELECT post_id 
+      FROM discourse_reactions_reactions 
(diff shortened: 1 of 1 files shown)


db/post_migrate/20220201162748_rename_thumbsup_reactions.rb
```diff
@@ -19,10 +19,31 @@ def up
...
-    DB.exec(<<~SQL, alias: alias_name, new_value: original_name)
+    has_both_reactions = DB.query_single(<<~SQL, alias: alias_name, new_value: original_name)
+      SELECT post_id 
+      FROM discourse_reactions_reactions 
+      WHERE reaction_value IN (:alias, :new_value) 
+      GROUP BY post_id 
+      HAVING COUNT(post_id) > 1
+    SQL
+
+    if has_both_reactions
```
eviltrout: Doesn't this need to check a rowcount or similar? query_single will return an empty array `[]` which would be truthy?

romanrizzi: Yes, good catch. Checking for `#present?` should be enough.

eviltrout (approved):

PR #143, opened 2022-02-16
FIX: remove duplicated code

KrisKotlarek:
Code to secure categories and access to specific topic is already in core in stable version: https://github.com/discourse/discourse/pull/14569/files

Those helper methods require builder to be an instance of `mini-sql`.

Tests were written in original PR https://github.com/discourse/discourse-reactions/pull/110/files

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -33,7 +33,7 @@ module DiscourseReactions
 
       reaction_users = DiscourseReactions::ReactionUser
         .joins(:reaction, :post)
-        .joins("INNER JOIN topics t on t.id = posts.topic_id")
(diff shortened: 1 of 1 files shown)


app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -32,8 +32,9 @@ def reactions_given
       raise Discourse::NotFound unless guardian.can_see_profile?(user)
 
       reaction_users = DiscourseReactions::ReactionUser
-        .joins(:reaction, post: :topic)
-        .joins("LEFT JOIN categories ON categories.id = topics.category_id")
+        .joins(:reaction, :post)
+        .joins("INNER JOIN topics t on t.id = posts.topic_id")
```
cvx:
```suggestion
        .joins("INNER JOIN topics t ON t.id = posts.topic_id")
```

cvx (approved):

PR #165, opened 2022-05-16
FIX: stores reference to expanded widget

jjaffeux:
It allows for a more accurate targeting of the previous popper to collapse when a different reaction panel is opened


nbianca (approved):
I usually do not like having such global variables, but I understand there is no other better way in this case.

jjaffeux:
> I usually do not like having such global variables, but I understand there is no other better way in this case.

Yes I really tried a lot of things before resorting to this, every other approaches I could think of were inconsistent :(

Also while it might leak a little bit, it should be quite limited.

PR #168, opened 2022-06-08
FIX: max reactions reached error message should be at par with core

arpitjalan:
https://meta.discourse.org/t/improve-copy-of-reaction-limit-reached-text/226718


arpitjalan:
Accompanying PR in core: https://github.com/discourse/discourse/pull/17037

cvx (approved):

arpitjalan:
Thanks for the review @CvX ❤️

PR #177, opened 2022-08-17
FIX: Protect from null value causing error

janzenisaac:


pmusaraj (approved):
Thanks. 

Approving so we can roll out the fix, but let's followup with a test, please.

PR #189, opened 2022-10-21
FIX: Reactions Like_Results query

TheJammiestDodger:
Update Reactions `like_results` query to scope to only the Like `post_action_type`


SamSaffron (approved):

SamSaffron:
Ideally we need a test on this, will add an internal note

PR #202, opened 2022-12-02
DEV: Apply Ember Octane Upgrades

keegangeorge:
This PR applies some ember upgrades to the plugin including things such as convert to native JS classes, apply angle bracket syntax, colocate components, use tracked getters/setters rather than computed properties, etc.

_Note this is the first set of changes (revolving around `/activity/reactions` `/notifications/reactions-received` routes), next set of changes (separate PR) will focus on updates to the widgets._


assets/javascripts/discourse/components/discourse-reactions-reaction-post.hbs
```diff
@@ -0,0 +1,59 @@
+<div
+  class="user-stream-item item {{if this.moderatorAction 'moderatorAction'}}"
```
keegangeorge: Unsure if converting the `classNameBindings` to this is the correct usage here.

assets/javascripts/discourse/routes/user-activity-reactions.js
```diff
@@ -1,13 +1,13 @@
-import DiscourseRoute from "discourse/routes/discourse";
+import Route from "@ember/routing/route";
 import CustomReaction from "../models/discourse-reactions-custom-reaction";
 
-export default DiscourseRoute.extend({
```
davidtaylorhq, line 4: I think it's probably safer to stick to `DiscourseRoute` - it has a few helpers that we use. You can still use the class-based syntax though 👌

assets/javascripts/discourse/components/discourse-reactions-reaction-post.hbs
davidtaylorhq: Yeah since there's only one conditional class, I think this is sensible. If the logic was mega-complex then it could've been refactored into getter, but there's no need for that here

assets/javascripts/discourse/controllers/user-activity-reactions.js
```diff
@@ -1,20 +1,22 @@
...
-  application: controller(),
-  beforeLikeId: null,
-  beforeReactionUserId: null,
+export default class UserActivityReactions extends Controller {
+  @controller application;
+  @service siteSettings;
+  canLoadMore = true;
+  loading = false;
+  beforeLikeId = null;
+  beforeReactionUserId = null;
```
davidtaylorhq: I suspect these will need to be `@tracked` so that setting them via `this.blah = ` actually triggers a rerender.

To make sure things are working, I'd recommend setting up your local environment with more than one page (25?) reactions, and check that infinite scrolling and the loading-spinner are working.

assets/javascripts/discourse/controllers/user-activity-reactions.js
```diff
@@ -61,18 +63,18 @@ export default Controller.extend({
...
         }
       })
       .finally(() => {
-        this.set("loading", false);
+        this.loading = false;
       });
-  },
+  }
 
   @observes("canLoadMore")
```
davidtaylorhq: Can you confirm that this is working correctly? I'm not sure whether `@observes` works under octane `@tracked`/autotracking

assets/javascripts/discourse/controllers/user-activity-reactions.js
```diff
@@ -61,18 +63,18 @@ export default Controller.extend({
...
   _showFooter() {
-    this.set("application.showFooter", !this.canLoadMore);
```
davidtaylorhq, line 76: `ApplicationController` hasn't been converted to use `@tracked` yet, so I think we've gotta stick with `.set()` here for now

davidtaylorhq (commented):
Looks great - nice work @keegangeorge!

We'll need to add a `.discourse-compatibility` entry to make sure that these new changes are not installed on `stable` versions of Discourse. (see [user-notes](https://github.com/discourse/discourse-user-notes/blob/main/.discourse-compatibility) for an example)

assets/javascripts/discourse/components/discourse-reactions-reaction-post.hbs
jjaffeux: you can use `{{concat-class "user-stream-item item" (if this.moderatorAction "moderatorAction")}}`

It avoids all the handling of white space

assets/javascripts/discourse/controllers/user-activity-reactions.js
keegangeorge: Interesting, yes it wasn't working correctly. I was getting new results loaded but the loading indicator wasn't appearing till I made the properties `@tracked`

assets/javascripts/discourse/controllers/user-activity-reactions.js
keegangeorge: Yup it wasn't working correctly, I've removed it in the latest commit and am updating `showFooter` differently, thanks!

keegangeorge:
Pushed some updates based on the review and feedback. May I have another look over please :)

assets/javascripts/discourse/controllers/user-activity-reactions.js
```diff
@@ -31,24 +38,25 @@ export default Controller.extend({
...
-      this.beforeReactionUserId = this._getLastIdFrom(reactionUsers);
+      this.beforeReactionUserId = this.#getLastIdFrom(reactionUsers);
     }
-  },
+  }
 
   @action
   loadMore() {
     if (!this.canLoadMore || this.loading || !this.reactionsUrl) {
+      this.set("application.showFooter", !this.canLoadMore);
```
davidtaylorhq, line 51: Much cleaner than the `@observes` approach, nice! 🙌

davidtaylorhq (approved):
Looks great, thanks @keegangeorge!

PR #214, opened 2023-03-01
UX: Adjust styling

jordanvidrine:
- Adjust styling of reactions to list to more closely resemble the reactions picker
- add min-height to allow scrolling for posts with many reactions

### After
image

### Before
image

image


plugin.rb
```diff
@@ -9,6 +9,7 @@
 
 enabled_site_setting :discourse_reactions_enabled
 
+register_asset "stylesheets/mixins/scrollbar.scss"
```
jjaffeux: where is this used?

jordanvidrine: Oops. I removed these and need to remove here as well.

jjaffeux (approved):

PR #217, opened 2023-03-23
UX: Re-use UserStreamItem from core

davidtaylorhq:
Previously this template file was copy/pasted/edited from core. Properly reusing the core component should help to ensure that any visual changes in core also apply to the reactions activity stream

Requires https://github.com/discourse/discourse/pull/20791

https://meta.discourse.org/t/user-stream-item-templates-needs-some-update/253476/9


pmusaraj (approved):
~~This also needs a `.discourse-compatibility` entry.~~ All good!

PR #224, opened 2023-07-17
FIX: prevents triggering a reaction while scrolling

jjaffeux:
It's a simple hack to mark a current touch sequence touchstart/touchend as invalid if the user has been moving finger (scrolling most likely) during start and end.

Not it will require https://github.com/discourse/discourse/pull/22635 to be merged for this to work, but it will just do nothing and not error if the commit is not present as `touchMove` won't be called in this case.

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-actions.js
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-actions.js
@@ -168,6 +168,8 @@ export default createWidget("discourse-reactions-actions", {
   },
 
   touchStart() {
+    this._validTouch = true;
+
     cancel(this._touchTimeout);
     if (this.capabilities.touch) {
       const root = document.getElementsByTagName("html")[0];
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/widgets/discourse-reactions-actions.js
```diff
@@ -168,6 +168,8 @@ export default createWidget("discourse-reactions-actions", {
   },
 
   touchStart() {
+    this._validTouch = true;
```
davidtaylorhq, line 171: Does this definitely work as expected? IIRC there's no guarantee that `this` will refer to the same Widget instance for each event (touch start/move/end).

jjaffeux: So I did check it works so, yes, it works.

And concerning the question itself, which is a good one. You don't get a new widget instance on these events, it's due to how hooks are configured:

```
  $(document).on("touchmove.discourse-widget", (e) => {
    nodeCallback(e.target, TOUCH_MOVE_ATTRIBUTE_NAME, (w) => w.touchMove(e), {
      rerender: false,
    });
  });
```

if I was changing to `rerender: true` I would get a new instance on each `touchmove` event.  Its something we can test by doing:


```javascript
touchMove() {
  console.log(guidFor(this));
}
```

With rerender false, we can move as much as we want the guid will stay the same.


--- 

At least this is my understanding, with widgets Im only half confident :p

davidtaylorhq (approved):

PR #229, opened 2023-08-07
DEV: Use new replacePostMenuButton api for like button

markvanlan:


jjaffeux (approved):

pmusaraj (commented):
Would `removePostMenuButton` and `addPostMenuButton` result in the same thing? Asking because we hesitate to add more widget-related API tools in core.

markvanlan:
> Would `removePostMenuButton` and `addPostMenuButton` result in the same thing? Asking because we hesitate to add more widget-related API tools in core.

This is specifically needed so that when plugins interact with eachother they play nice.. We need this so that we don't have to hide reactions with CSS `https://github.com/discourse/discourse-post-voting/pull/153/files`.. There is no way for the post-voting plugin to tell core to render or NOT to render reactions, so replacing the like and then showing or hiding that in the post-voting plugin works.

PR #230, opened 2023-08-10
FIX: Display correct title text for unauthenticated users

Drenmi:
[Meta](https://meta.discourse.org/t/wrong-tooltip-on-like-button-for-non-registered-user/273896).

### What was the issue?

Unauthenticated users would incorrectly see this title text when hovering the reaction button:

Screenshot 2023-08-10 at 11 33 05 AM

### How does this fix it?

This change adds a new translation key and uses it when user is authenticated:

Screenshot 2023-08-10 at 11 32 43 AM

### Should there be a test for this?

I am on the fence. Should QUnit/system tests go to the granularity of title texts? I'm happy to add it if there's a compelling case made. WDYT?

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-reaction-button.js
@@ -50,6 +50,12 @@ export default createWidget("discourse-reactions-reaction-button", {
   },
 
   buildAttributes(attrs) {
+    if (!this.currentUser) {
+      return {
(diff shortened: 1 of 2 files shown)


martin-brennan (approved):
Looks nice 👍 Personally I think this is too small for a system test, especially since you have to hover to even see it and if it shows the wrong message it's not a big deal. It's a widget too so it's unlikely to be easy to test, and I know for sure that Joffrey is currently converting all this to Glimmer components, so I wouldn't worry too much about tests for this.

PR #235, opened 2023-09-04
FIX: Adding DATE casts to the Reactions dashboard report

TheJammiestDodger:
The Reactions dashboard report did not include data from the day of the end_date. Casting the respective created_at values as DATE should now include this.

/t/73842


nattsw (approved):
Let's get this merged then write a test after 👍

PR #242, opened 2023-09-12
DEV: Update linting setup

cvx:


Gemfile
```diff
@@ -5,4 +5,5 @@ source "https://rubygems.org"
 group :development do
   gem "rubocop-discourse"
   gem "syntax_tree"
+  gem "syntax_tree-disable_ternary"
```
davidtaylorhq, line 8: This is built-in to syntax_tree now, so we should be able to drop the extra gem and use `--plugins=plugin/disable_auto_ternary`

davidtaylorhq (approved):

Gemfile
cvx: (note: we decided to wait for the new release of stree)

davidtaylorhq: (scrap that, the official version isn't released yet)

PR #245, opened 2023-09-13
DEV: Pluralized some translation strings

tgxworld:
Why this change?

This allows other locales to customize the pluralization specific to the
locales even though it is not required for the `en` locale.


dbattersby (approved):

tgxworld:
Thank you @dbattersby

PR #250, opened 2023-10-19
FIX: update post id for reactions when post is moved

dbattersby:
Currently all emoji reactions on a post are lost when posts are moved to a new topic.

This change updates the reaction data post id to the new post.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -344,21 +344,29 @@ after_initialize do
   register_notification_consolidation_plan(reacted_by_two_users)
   register_notification_consolidation_plan(consolidated_reactions)
 
-  on(:post_moved) do |post, original_topic_id, original_post_id|
-    next if original_post_id.nil?
-    params = { old_post_id: original_post_id, new_post_id: post.id }
+  on(:first_post_moved) do |target_post, original_post|
+    reaction_cols = (DiscourseReactions::Reaction.column_names - %w[id post_id]).join(", ")
+    reaction_user_cols =
+      (DiscourseReactions::ReactionUser.column_names - %w[id post_id reaction_id]).join(", ")
 
     ActiveRecord::Base.transaction do
-      DB.exec(<<~SQL, params)
-        UPDATE discourse_reactions_reactions
-        SET post_id = :new_post_id
-        WHERE post_id = :old_post_id
+      DB.exec <<~SQL
+      INSERT INTO discourse_reactions_reactions(post_id, #{reaction_cols})
+      SELECT #{target_post.id}, #{reaction_cols}
+      FROM discourse_reactions_reactions
+      WHERE post_id = #{original_post.id}
       SQL
 
-      DB.exec(<<~SQL, params)
-        UPDATE discourse_reactions_reaction_users
-        SET post_id = :new_post_id
-        WHERE post_id = :old_post_id
+      reaction_id =
+        DB.query_single(
+          "SELECT id FROM discourse_reactions_reactions ORDER BY id DESC LIMIT 1",
+        ).first
+
+      DB.exec <<~SQL
+      INSERT INTO discourse_reactions_reaction_users(post_id, reaction_id, #{reaction_user_cols})
+      SELECT #{target_post.id}, #{reaction_id}, #{reaction_user_cols}
+      FROM discourse_reactions_reaction_users
+      WHERE post_id = #{original_post.id}
       SQL
     end
   end
--- a/plugin.rb
+++ b/plugin.rb
@@ -345,29 +345,24 @@ after_initialize do
   register_notification_consolidation_plan(consolidated_reactions)
 
   on(:first_post_moved) do |target_post, original_post|
-    reaction_cols = (DiscourseReactions::Reaction.column_names - %w[id post_id]).join(", ")
-    reaction_user_cols =
-      (DiscourseReactions::ReactionUser.column_names - %w[id post_id reaction_id]).join(", ")
-
+    id_map = {}
     ActiveRecord::Base.transaction do
-      DB.exec <<~SQL
-      INSERT INTO discourse_reactions_reactions(post_id, #{reaction_cols})
-      SELECT #{target_post.id}, #{reaction_cols}
-      FROM discourse_reactions_reactions
-      WHERE post_id = #{original_post.id}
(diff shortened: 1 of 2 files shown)


tgxworld:
:grin: Is there a reason why didn't add a test here?

dbattersby:
I had meant to add a test for this but must have missed it, let me correct that 😄

plugin.rb
```diff
@@ -343,4 +343,23 @@ class Engine < ::Rails::Engine
...
+        UPDATE discourse_reactions_reactions
+        SET post_id = :new_post_id
+        WHERE post_id = :old_post_id
+      SQL
+
+      DB.exec(<<~SQL, params)
+        UPDATE discourse_reactions_reaction_users
+        SET post_id = :new_post_id
+        WHERE post_id = :old_post_id
+      SQL
```
tgxworld: I don't think this is right because the original post which still exists will lose its reactions in this case. In Discourse core when we move the first post, it actually copies all the existing `PostActions` so I think we need to follow that pattern as well.

https://github.com/discourse/discourse/blob/81c2d38cea7c805cca4dfc7f4245506554991f12/app/models/post_mover.rb#L261-L263

dbattersby: I was just thinking about this earlier. My initial thinking here was that we are moving the post so these reactions are orphaned but for the first post it is an entirely new post being created, so we can't move them in this way without removing them from the original post.

I'll refactor this to work similar to how `PostAction.copy` works.

dbattersby: I've updated this to copy the reactions and insert them for the new post (only for first post in topic).

plugin.rb
```diff
@@ -343,4 +343,31 @@ class Engine < ::Rails::Engine
...
+      INSERT INTO discourse_reactions_reactions(post_id, #{reaction_cols})
+      SELECT #{target_post.id}, #{reaction_cols}
+      FROM discourse_reactions_reactions
+      WHERE post_id = #{original_post.id}
+      SQL
+
+      reaction_id =
+        DB.query_single(
+          "SELECT id FROM discourse_reactions_reactions ORDER BY id DESC LIMIT 1",
+        ).first
```
tgxworld: Sorry I'm not sure I understand this because this is just fetching the last reaction created. What if there are multiple reactions inserted from the SQL query above?

dbattersby: You are right, this turned out to be more tricky than initially expected.

I think it would easily get very unreadable if I continued with a raw query, so I decided to simplify this by using Active Record and merging the values then `insert_all`. I think it is more readable and works correctly now.

The test has also been updated to account for multiple reactions on the same post.

Edit: I think the performance cost between raw and AR is minimal as we are batch inserting.

spec/models/post_mover_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
tgxworld: This line line is [not necessary](https://github.com/discourse/discourse/blob/a2000a3559eef9ef17d83ee79cf9a3319eb5df24/.rspec#L1C4-L1C4).

spec/models/post_mover_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+require_relative "../fabricators/reaction_fabricator.rb"
+require_relative "../fabricators/reaction_user_fabricator.rb"
```
tgxworld:
```suggestion
```

This two lines are not necessary too since we [load plugins fabricators in core](https://github.com/discourse/discourse/blob/a2000a3559eef9ef17d83ee79cf9a3319eb5df24/spec/rails_helper.rb#L92).

spec/models/post_mover_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  fab!(:reaction_3) { Fabricate(:reaction, post: post_2, reaction_value: "heart") }
+  fab!(:reaction_4) { Fabricate(:reaction, post: post_2, reaction_value: "wave") }
+  fab!(:user_reaction_1) { Fabricate(:reaction_user, reaction: reaction_1, post: post_1) }
+  fab!(:user_reaction_2) { Fabricate(:reaction_user, reaction: reaction_2, post: post_1) }
+  fab!(:user_reaction_3) { Fabricate(:reaction_user, reaction: reaction_3, post: post_2) }
+  fab!(:user_reaction_4) { Fabricate(:reaction_user, reaction: reaction_4, post: post_2) }
+
+  before { SiteSetting.discourse_reactions_enabled = true }
+
+  it "new post has topic's first post reactions (OP)" do
```
tgxworld: Minor one but rspec descriptions are supposed to read like a proper sentence. Perhaps something like this? 

```suggestion
  it "should add old post's reactions to new post when a topic's first post is moved" do
```

spec/models/post_mover_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    expect(post_1.reactions).to include(reaction_1)
```
tgxworld:
```suggestion
    expect(post_1.reactions).to contain_exactly(reaction_1, reaction_2)
```

spec/models/post_mover_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    expect(reaction_emojis.count).to eq(2)
+    expect(reaction_emojis).to match_array([reaction_1.reaction_value, reaction_2.reaction_value])
+
+    reaction_user_ids = new_post.reactions_user.pluck(:user_id)
+    expect(reaction_user_ids.count).to eq(2)
+    expect(reaction_user_ids).to match_array([user_reaction_1.user_id, user_reaction_2.user_id])
+  end
+
+  it "moved post still has existing reactions" do
```
tgxworld:
```suggestion
  it "should retain existing reactions after moving a post" do
```

spec/models/post_mover_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    expect(post_2.reactions).to include(reaction_3)
```
tgxworld:
```suggestion
    expect(post_2.reactions).to contain_exactly(reaction_3, reaction_4)
```

tgxworld (approved):

PR #255, opened 2023-11-06
FIX: missing reaction user for heart emoji breaks post mover

dbattersby:
When a user reacts to a post it creates an entry for both the `discourse_reactions_reactions` and `discourse_reactions_reaction_users` tables.

However, when a post is liked using the :heart: emoji it still creates a `discourse_reactions_reactions` entry but stores user data within `post_actions` in core and doesn't create a `discourse_reactions_reaction_users` entry.

This change adds an extra check to skip adding `DiscourseReactions::ReactionUser` data if none is found.

The new test recreates the logic outlined above to assume that the post has a reaction and not the reaction user when it is liked, and should still work correctly.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -353,11 +353,14 @@ after_initialize do
 
       reactions_attributes =
         reactions.map { |reaction| reaction.attributes.except("id").merge(post_id: target_post.id) }
+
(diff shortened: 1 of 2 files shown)


jjaffeux (approved):
LGTM

PR #267, opened 2024-01-29
FEATURE: Count all reactions as likes with exceptions controlled by a site setting

martin-brennan:
This commit changes the reaction paradigm to instead count all reactions as
Likes (via a `PostAction` record) rather than only considering a single specific
reaction emoji as a Like equivalent.

A new site setting, `discourse_reactions_excluded_from_like`, is introduced defaulting
only to have :-1: . Any emojis on this list will **not** create a `PostAction` record that
represents a Like.

We now create a `PostAction` record along with the existing `ReactionUser` record
for reactions which are not excluded from Like, to help with history correction if
`discourse_reactions_excluded_from_like` changes.

For existing settings:

* `discourse_reactions_reaction_for_like` is still used, since we need an easy way to map
  regular Likes to reactions if the plugin is enabled/disabled. This cannot be set to anything
  in `discourse_reactions_excluded_from_like`. If this is changed, history will not get changed.
  Any reaction using this setting will only make a `PostAction` record.
* `discourse_reactions_like_icon` is still used, since that is purely a visual setting.

Also included a job that performs data consistency operations whenever
`discourse_reactions_excluded_from_like` is changed, to create or trash
relevant `PostAction` records, and recover already trashed ones as needed.
Also `UserAction` records for `LIKE` and `WAS_LIKED` are created and deleted
as needed.

For now `discourse_reactions_like_sync_enabled` controls whether we create/trash
`PostAction` and `UserAction` records when the `discourse_reactions_excluded_from_like`
is changed. We are not including a migration until this is running in the wild for a few weeks.


app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -136,19 +135,24 @@ def reactions_received
   end
 
   def post_reactions_users
-    id = params.require(:id).to_i
+    params.require(:id).to_i
```
SamSaffron, line 138: what is the point of `to_i` here?

lib/tasks/reactions.rake
```diff
@@ -148,7 +148,6 @@ task "reactions:generate", %i[post_id reactions_count reaction] => [:environment
     DiscourseReactions::ReactionManager.new(
       reaction_value: reaction,
       user: user,
-      guardian: Guardian.new(user),
```
SamSaffron, line 151: what is the context for removing the guardian here?

martin-brennan: `user` always has a guardian attached via `user.guardian`, it's unnecessary to re-initialize it all the time; I see this problem in core a lot too

app/controllers/discourse_reactions/custom_reactions_controller.rb
martin-brennan: No idea TBH I just removed the assignment to get rid of the rubocop warning :) will remove to_i as well

plugin.rb
```diff
@@ -144,12 +170,21 @@ class Engine < ::Rails::Engine
   end
 
   add_to_serializer(:post, :current_user_used_main_reaction) do
-    return false unless scope.user.present?
+    return false if scope.is_anonymous?
```
SamSaffron, line 173: you know how much I prefer this :hugs:

plugin.rb
```diff
@@ -372,4 +402,11 @@ class Engine < ::Rails::Engine
       DiscourseReactions::ReactionUser.insert_all(reaction_users_attributes)
     end
   end
+
+  on(:site_setting_changed) do |name, old_value, new_value|
+    if name == :discourse_reactions_excluded_from_like
```
SamSaffron: nothing to change here, but you can see how non ideal this api is in the long term... any site setting that changes always call the block, nothing to change now, but it is interesting.

app/jobs/regular/discourse_reactions/post_action_synchronizer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
```
SamSaffron, line 2: hmmm how do we kick off the "first" sync? Do we just create a scheduled job that checks if we are behind on a sync?

db/migrate/20240130013701_create_post_action_reaction_user_historical_records.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class CreatePostActionReactionUserHistoricalRecords < ActiveRecord::Migration[7.0]
+  def up
```
SamSaffron: oh ... the first sync is here... I see...

plugin.rb
```diff
@@ -73,25 +78,45 @@ class Engine < ::Rails::Engine
   end
 
   add_to_serializer(:post, :reactions) do
```
SamSaffron, line 80: the post serializer was modified a fair bit, is it covered? a "GET" of posts that looks at the reactions?

SamSaffron:
Overall looking very good, :+1:  approving

SamSaffron (approved):

plugin.rb
martin-brennan: I ruthlessly prune `unless` everywhere I go 😆

plugin.rb
martin-brennan: Yeah I think we have discussed this previously for something else? I can't find it on dev but it definitely sounds familiar.

plugin.rb
martin-brennan: Yeah there are already tests for the serializer and for requests. I will go and double check though and make sure I've got the coverage needed.

db/migrate/20240130013701_create_post_action_reaction_user_historical_records.rb
martin-brennan: Yeah I thought this would be best. It did just occur to me though that maybe this would be better as a post migration.

app/jobs/regular/discourse_reactions/post_action_synchronizer.rb
```diff
@@ -0,0 +1,11 @@
...
+module Jobs
+  module DiscourseReactions
+    class PostActionSynchronizer < ::Jobs::Base
+      def execute(args = {})
+        ::DiscourseReactions::ReactionPostActionSynchronizer.sync!
```
SamSaffron, line 12: what triggers this, should this not be a "every 1 day" or something like that?

SamSaffron: oh I see, setting change... so initial sync will never fire... we need to make this a regular job instead and check if it "needs to synchronize"

martin-brennan: No it's not necessary, this only is triggered when the `discourse_reactions_excluded_from_like` setting changes for backfilling:

https://github.com/discourse/discourse-reactions/blob/e6f3ca690e6e9d67bfe6350421424fe5f5c27cb7/plugin.rb#L406-L411

The `ReactionManager` is used for general usage when you are clicking in the UI. Internally, we will set `discourse_reactions_like_sync_enabled` to `true` then manually trigger this job for testing.

PR #270, opened 2024-02-13
FIX: Handle null post.user_id for UserAction sync

martin-brennan:
There can be rare cases where a post user_id can be
NULL, this handles that case and also handles conflicts
when UserAction records are inserted that match existing
ones.


dbattersby (approved):
Looks good to me 👍

PR #274, opened 2024-02-16
FIX: Hovering reaction icon causing error would flood AJAX requests

martin-brennan:
If a user hovered over a reaction to see a list of users,
mostly a problem when anon browsing (see
https://github.com/discourse/discourse/commit/5935148bd863ea80865cee1049b3a44f3fbf3cc8),
we didn't have a throttle when erroring, so extra hovers would cause
error messages.

This fixes the issue by adding a throttle, and also adds some general
reactions system testing for this area.


assets/javascripts/discourse/widgets/discourse-reactions-list-emoji.js
```diff
@@ -109,4 +123,10 @@ export default createWidget("discourse-reactions-list-emoji", {
       }
     });
   },
+
+  _allowHover() {
```
martin-brennan, line 120: Yes this is not ideal, but there is seemingly no way to get rid of `(hover: none)` in chrome headless. Maybe in time we could globally set `window.isSystemTest` in system tests, like how we have `isTesting()` available from ember

assets/javascripts/discourse/widgets/discourse-reactions-list-emoji.js
```diff
@@ -16,11 +16,25 @@ export default createWidget("discourse-reactions-list-emoji", {
...
-        this.callWidgetFunction("getUsers", this.attrs.reaction.id);
+      if (
+        !this.attrs.users?.length &&
+        !this.debounceLoadReactions &&
+        !this.loadingReactions
+      ) {
+        this.loadingReactions = true;
+        this.callWidgetFunction("getUsers", this.attrs.reaction.id)
+          .catch(() => {
+            this.debounceLoadReactions = true;
```
Drenmi: Any chance we can use [`#debounce`](https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/debounce) from `@ember/runloop`? 🤔

assets/javascripts/discourse/widgets/discourse-reactions-list-emoji.js
Drenmi: > Maybe in time we could globally set `window.isSystemTest` [...]

👍

Drenmi (approved):
LGTM! 🚀

assets/javascripts/discourse/widgets/discourse-reactions-list-emoji.js
martin-brennan: Yes this works! Using that instead :)

PR #275, opened 2024-02-19
FIX: Do not show users who Reacted under post ... menu

martin-brennan:
There is a "show-more-actions" button represented by ...
at the bottom of each post. This shows the users who have
liked the post. However, now that reactions also count as likes,
we don't want to show the users who reacted to the post here,
only the ones who liked it with main_reaction_id.

We can use the plugin modifier introduced in core in
https://github.com/discourse/discourse/pull/25740 for this.

image

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -385,6 +385,23 @@ after_initialize do
     builder.where("discourse_reactions_reaction_users.id IS NULL")
   end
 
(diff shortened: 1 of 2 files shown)


Drenmi (approved):
Cool!

I assume the failing test will pass once the core PR is merged. 👍

PR #278, opened 2024-02-23
FIX: view activity reactions for other users

dbattersby:
This change restores access to the reaction activity of other users when the current user is logged in.


nattsw (approved):
Nice tests

PR #283, opened 2024-03-21
DEV: Additional backfilling for like sync

martin-brennan:
Adding more Like backfilling for these tables and columns:

1. TopicUser.liked
2. Post.like_count
3. Topic.like_count
4. UserStat.likes_given
5. UserStat.likes_received
6. GivenDailyLike, which is used for the Crazy In Love badge, and is incremented/decremented whenever you like/unlike something

This also adds `DiscourseReactions::MigrationReport` which provides output
like this to see if things work correctly:

```
Reaction migration report:
------------------------------------------------------------

main_reaction_id:                       heart
discourse_reactions_like_sync_enabled:  true
discourse_reactions_enabled_reactions:  +1|laughing|open_mouth|clap|confetti_ball|hugs
discourse_reactions_excluded_from_like: +1|laughing|open_mouth|clap|confetti_ball|hugs

PostAction likes:        682006 (-8078)
UserActions.liked:       681908 (-8077)
UserActions.was_liked:   681908 (-8077)
UserStat.likes_given:    669055 (-8077)
UserStat.likes_received: 681176 (-8077)
Post.like_count:         683526 (-8078)
Topic.like_count:        683526 (-8078)
TopicUser.liked:         353309 (-1426)
Badge count:             38998 (no change)
Given daily like total:  682834 (-8077)
Reactions:               48284 (no change)
 -> +1: 7951
 -> laughing: 7908
 -> open_mouth: 8012
 -> clap: 8128
 -> confetti_ball: 8207
 -> hugs: 8078
```


app/services/discourse_reactions/reaction_like_synchronizer.rb
```diff
@@ -0,0 +1,439 @@
+# frozen_string_literal: true
+
+module DiscourseReactions
+  class ReactionLikeSynchronizer
+    def self.sync!
+      return if !SiteSetting.discourse_reactions_like_sync_enabled
+
+      excluded_from_like = SiteSetting.discourse_reactions_excluded_from_like.to_s.split("|")
```
KrisKotlarek: Can this be `class_attribute` so later we don't need to pass it methods like `create_missing_post_actions` or ` recover_trashed_post_actions` by hand?
I think it is a configuration value which is accessed in many places

app/services/discourse_reactions/reaction_like_synchronizer.rb
```diff
@@ -0,0 +1,439 @@
...
+    # excluded_from_like, and create a PostAction record for each.
+    def self.create_missing_post_actions(excluded_from_like)
+      sql_query = <<~SQL
+        INSERT INTO post_actions(
+          post_id, user_id, post_action_type_id, created_at, updated_at
+        )
+        SELECT ru.post_id, ru.user_id, :pa_like, ru.created_at, ru.updated_at
+        FROM discourse_reactions_reaction_users ru
+        INNER JOIN discourse_reactions_reactions
+          ON discourse_reactions_reactions.id = ru.reaction_id
```
KrisKotlarek, line 82: Why do we need that inner join for `discourse_reactions_reactions`? To ensure that reaction is still available?

If yes, maybe we shouldn't do that. If someone reacted in the past, when a reaction was available, it should **count** even if now configuration of the forum changed as it was a valid "like".

app/services/discourse_reactions/reaction_like_synchronizer.rb
```diff
@@ -0,0 +1,439 @@
...
+            AND post_actions.post_action_type_id = :like
+            AND discourse_reactions_reactions.reaction_value IN (:excluded_from_like)
+          RETURNING post_actions.id, post_actions.post_id, post_actions.user_id
+        )
+
+        DELETE FROM user_actions
+        USING deleted_post_actions
+        WHERE user_actions.target_post_id = deleted_post_actions.post_id
+        AND user_actions.acting_user_id = deleted_post_actions.user_id
+        AND user_actions.action_type IN (:ua_like, :ua_was_liked)
```
KrisKotlarek: You are deleting user actions here so I think that method `delete_excluded_user_actions` is not needed anymore. Or keep `delete_excluded_user_actions` and remove block above.

app/services/discourse_reactions/reaction_like_synchronizer.rb
```diff
@@ -0,0 +1,439 @@
...
+      #     LEFT JOIN post_actions ON post_actions.post_id = posts.id
+      #       AND post_actions.post_action_type_id = :like
+      #     WHERE post_actions.id IN (:all_affected_post_action_ids)
+      #     GROUP BY posts.id
+      #   )
+      #   UPDATE posts
+      #   SET like_count = like_counts.like_count
+      #   FROM like_counts
+      #   WHERE posts.id = like_counts.post_id
+      # SQL
```
KrisKotlarek: This looks like some leftovers

lib/discourse_reactions/migration_report.rb
```diff
@@ -0,0 +1,87 @@
...
+        user_actions_was_liked: UserAction.where(action_type: UserAction::WAS_LIKED).count,
+        post_action_likes: PostAction.where(post_action_type_id: PostActionType.types[:like]).count,
+        post_like_count_total: Post.sum(:like_count),
+        topic_like_count_total: Topic.sum(:like_count),
+        user_stat_likes_given_total: UserStat.sum(:likes_given),
+        user_stat_likes_received_total: UserStat.sum(:likes_received),
+        given_daily_like_total: GivenDailyLike.sum(:likes_given),
+        badge_count: UserBadge.count,
+        reactions_total: DiscourseReactions::ReactionUser.count,
+      }
```
KrisKotlarek, line 19: Similar to previous comment, report data can be `class_attribute`

app/services/discourse_reactions/reaction_like_synchronizer.rb
martin-brennan: Yes good point -- you know I could just make it a class that is instanced actually, it would make this easier. Will do this.

app/services/discourse_reactions/reaction_like_synchronizer.rb
martin-brennan: Oh nice catch, I will delete this block

app/services/discourse_reactions/reaction_like_synchronizer.rb
martin-brennan: Sorry :grimacing: Removing!

lib/discourse_reactions/migration_report.rb
martin-brennan: I would prefer to leave this one as is -- it will be much more helpful for debugging this way

app/services/discourse_reactions/reaction_like_synchronizer.rb
martin-brennan: Sorry I am not following here. `discourse_reaction_reactions` is an aggregate table, it only stores the distinct reaction values and count for a specific post:

```
# Table name: discourse_reactions_reactions
#
#  id                   :bigint           not null, primary key
#  post_id              :integer
#  reaction_type        :integer
#  reaction_value       :string
#  reaction_users_count :integer
#  created_at           :datetime         not null
#  updated_at           :datetime         not null
```

I don't think we can assume intent on past reactions, we have to only rely on the current state of `excluded_from_like` . Maybe I am misinterpreting your meaning?

KrisKotlarek: You are right, I read it incorrectly

KrisKotlarek (approved):

PR #288, opened 2024-04-15
FIX: ensures activity/reactions doesnt show deleted

jjaffeux:
This fix ensures our query doesn't retrieve deleted posts which would cause an exception on the page when the js code would try to access the associated post which would be null in this case.


martin-brennan (approved):
Thanks for this, dumb that I missed this case :facepalm:

PR #289, opened 2024-05-15
FIX: Reactions doubled up as like if no longer enabled

martin-brennan:
This commit fixes this scenario:

1. A user reacts to a post (+1), and their reaction also counts as a like
1. Another user sees the post and sees only +1
1. The admin changes SiteSetting.discourse_reactions_enabled_reactions
   to no longer enable the +1 reaction
1. The user now sees both +1 and a heart on the post

The fix makes it so the post still only shows a +1 even
though that is no longer an "enabled" reaction since it is
historical data.

It also fixes up some other discrepencies with like counts
and reaction counts based on changing settings after the fact,
like changing the `main_reaction_id` (which is definitely not advised).

### DIFF
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -172,7 +172,7 @@ class DiscourseReactions::CustomReactionsController < ApplicationController
 
       # Filter out likes for reactions that are not longer enabled,
       # which match up to a ReactionUser in historical data.
-      historical_likes =
+      historical_reaction_likes =
         likes
           .joins(
             "LEFT JOIN discourse_reactions_reaction_users ON discourse_reactions_reaction_users.user_id = post_actions.user_id AND discourse_reactions_reaction_users.post_id = post_actions.post_id",
@@ -185,7 +185,7 @@ class DiscourseReactions::CustomReactionsController < ApplicationController
             valid_reactions: DiscourseReactions::Reaction.valid_reactions.to_a,
           )
 
-      likes = likes.where.not(id: historical_likes.pluck(&:id))
+      likes = likes.where.not(id: historical_reaction_likes.select(&:id))
     end
 
     if likes.present?
--- a/app/controllers/discourse_reactions/custom_reactions_controller.rb
+++ b/app/controllers/discourse_reactions/custom_reactions_controller.rb
@@ -162,13 +162,30 @@ class DiscourseReactions::CustomReactionsController < ApplicationController
 
     if !reaction_value || reaction_value == DiscourseReactions::Reaction.main_reaction_id
       # We only want to get likes that don't have an associated ReactionUser
-      # record, which count as a like or will be double ups for main_reaction_id.
+      # record, which count as a like or there will be double ups for main_reaction_id.
       likes =
         post.post_actions.where(
           DiscourseReactions::PostActionExtension.filter_reaction_likes_sql,
           like: PostActionType.types[:like],
           valid_reactions: DiscourseReactions::Reaction.valid_reactions.to_a,
         )
+
+      # Filter out likes for reactions that are not longer enabled,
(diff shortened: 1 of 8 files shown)


app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -162,13 +162,30 @@ def post_reactions_users
...
+          )
+          .joins(
+            "LEFT JOIN discourse_reactions_reactions ON discourse_reactions_reactions.id = discourse_reactions_reaction_users.reaction_id",
+          )
+          .where(
+            "discourse_reactions_reactions.reaction_value NOT IN (:valid_reactions)",
+            valid_reactions: DiscourseReactions::Reaction.valid_reactions.to_a,
+          )
+
+      likes = likes.where.not(id: historical_likes.pluck(&:id))
```
Drenmi: Think we can do this to avoid one SQL query to load the IDs into Ruby.

```suggestion
      likes = likes.where.not(id: historical_likes.select(:id))
```

plugin.rb
```diff
@@ -93,7 +93,11 @@ class Engine < ::Rails::Engine
           count: reaction.reaction_users_count,
         }
 
-        if DiscourseReactions::Reaction.reactions_counting_as_like.include?(reaction.reaction_value)
+        # NOTE: It does not matter if the reaction is currently an enabled one,
+        # we need to handle historical data here too so we don't see double-ups in the UI.
+        if !DiscourseReactions::Reaction.reactions_excluded_from_like.include?(
```
Drenmi, line 98: Seeing `DiscourseReactions::Reaction.reactions_excluded_from_like.include?` in multiple places. Maybe should be a method on `DiscourseReactions::Reaction`?

plugin.rb
```diff
@@ -110,15 +114,25 @@ class Engine < ::Rails::Engine
           # Get rid of any PostAction records that match up to a ReactionUser
           # that is NOT main_reaction_id and is NOT excluded, otherwise we double
           # up on the count/reaction shown in the UI.
-          reaction_users_counting_as_like.find { |ru| ru.user_id == post_action.user_id }.present?
+          reaction_users_counting_as_like
+            .find { |ru| ru.user_id == post_action.user_id }
+            .present? ||
```
Drenmi:
```suggestion
            .any? { |ru| ru.user_id == post_action.user_id } ||
```

Drenmi (approved):
> The fix makes it so the post still only shows a +1 even
though that is no longer an "enabled" reaction since it is
historical data.

Agree! I think "the ink is dry" is a good principle to follow here.

LGTM with some small suggestions. 💪

app/controllers/discourse_reactions/custom_reactions_controller.rb
martin-brennan: Ah yeah I forget about this, it's pretty trippy how it works. IIRC I only learned about this from Daniel not long ago

plugin.rb
martin-brennan: Hmm I think it's fine, it's only 2 places and just an array check

app/controllers/discourse_reactions/custom_reactions_controller.rb
```diff
@@ -162,13 +162,30 @@ def post_reactions_users
...
+          )
+          .joins(
+            "LEFT JOIN discourse_reactions_reactions ON discourse_reactions_reactions.id = discourse_reactions_reaction_users.reaction_id",
+          )
+          .where(
+            "discourse_reactions_reactions.reaction_value NOT IN (:valid_reactions)",
+            valid_reactions: DiscourseReactions::Reaction.valid_reactions.to_a,
+          )
+
+      likes = likes.where.not(id: historical_reaction_likes.select(&:id))
```
Drenmi, line 188: Need to remove the `&` for it to work without the extra query. 😬 

```ruby
likes.where.not(id: historical_reaction_likes.select(:id))
````

PR #292, opened 2024-05-20
FIX: N1 in post action extension

martin-brennan:
Followup 0b4b186409f2704d29c459c8f2555817d314b380

The previous commit caused an N1 because I was calling
post_action.reaction_user in a loop, which was making
a query for every Like PostAction attached to a Post.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -104,6 +104,20 @@ after_initialize do
 
     reaction_users_counting_as_like.flatten!
 
+    # Used to remove historical like data in the filter below.
+    previous_main_reaction_like_ids =
+      object
+        .post_actions
+        .where(
+          "post_actions.post_action_type_id = :like AND
+            post_actions.post_id IN (
+              #{DiscourseReactions::PostActionExtension.post_action_with_reaction_user_sql}
+            )",
+          like: PostActionType.types[:like],
+          valid_reactions: [DiscourseReactions::Reaction.main_reaction_id],
+        )
+        .pluck(:id)
+
     likes =
       object
         .post_actions
@@ -119,10 +133,7 @@ after_initialize do
 
           # Also get rid of any PostAction records that match up to a ReactionUser
           # that is now the main_reaction_id and has historical data.
-          is_previously_enabled_reaction =
-            post_action.reaction_user.present? &&
-              post_action.reaction_user.reaction.reaction_value ==
-                DiscourseReactions::Reaction.main_reaction_id
+          is_previously_enabled_reaction = previous_main_reaction_like_ids.include?(post_action.id)
 
           is_reaction_like_duplicate || is_previously_enabled_reaction
         end
--- a/plugin.rb
+++ b/plugin.rb
@@ -104,39 +104,24 @@ after_initialize do
 
     reaction_users_counting_as_like.flatten!
 
-    # Used to remove historical like data in the filter below.
-    previous_main_reaction_like_ids =
-      object
-        .post_actions
-        .where(
-          "post_actions.post_action_type_id = :like AND
-            post_actions.post_id IN (
-              #{DiscourseReactions::PostActionExtension.post_action_with_reaction_user_sql}
-            )",
-          like: PostActionType.types[:like],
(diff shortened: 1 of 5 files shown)


plugin.rb
```diff
@@ -104,6 +104,20 @@ class Engine < ::Rails::Engine
 
     reaction_users_counting_as_like.flatten!
 
+    # Used to remove historical like data in the filter below.
+    previous_main_reaction_like_ids =
```
martin-brennan: We can potentially add a spec using `track_sql_queries` for this, but since this is one of many `add_to_serializer` modifications to the `PostSerializer`, it's not really possible to do a clean before/after comparison without some more extensive refactoring, which I do not think is worth it at this time. Confirmed that the N1 is now gone locally.

ZogStriP (approved):

plugin.rb
```diff
@@ -104,6 +104,20 @@ class Engine < ::Rails::Engine
...
+      object
+        .post_actions
+        .where(
+          "post_actions.post_action_type_id = :like AND
+            post_actions.post_id IN (
+              #{DiscourseReactions::PostActionExtension.post_action_with_reaction_user_sql}
+            )",
+          like: PostActionType.types[:like],
+          valid_reactions: [DiscourseReactions::Reaction.main_reaction_id],
+        )
```
tgxworld: Since we are rejecting all `post_actions` that meet this condition eventually, I'm wondering if we can just add this filter to the query below.

```
      object
        .post_actions
        .where(
          "post_actions.post_action_type_id = :like AND
            post_actions.post_id IN (
              #{DiscourseReactions::PostActionExtension.post_action_with_reaction_user_sql}
            )",
          like: PostActionType.types[:like],
          valid_reactions: [DiscourseReactions::Reaction.main_reaction_id],
        )
        .select do |post_action|
         ...
       end
       .reject do |post_action|
        ...
       end
```

jjaffeux (approved):

plugin.rb
```diff
@@ -104,6 +104,20 @@ class Engine < ::Rails::Engine
...
+        .pluck(:id)
```
tgxworld: Also, if this is being added to the `PostSerializer`, it will still cause an N+1 if we are serializing an array of `Post` records.

martin-brennan: Ah yeah you're right :/ It's just a "less bad" N1. Will find a way to fix before  merging.

plugin.rb
martin-brennan: Oh yeah that might work -- checking.

plugin.rb
```diff
@@ -105,27 +105,27 @@ class Engine < ::Rails::Engine
...
+            .post_actions_with_reaction_users
+            &.find do |pa|
+              pa.id == post_action.id &&
+                pa.reaction_user&.reaction&.reaction_value ==
+                  DiscourseReactions::Reaction.main_reaction_id
+            end
+            .present?
+
+        is_reaction_like_duplicate || is_previously_enabled_reaction
+      end
```
tgxworld:
```suggestion
      object.post_actions.reject do |post_action|
        # Get rid of any PostAction records that match up to a ReactionUser
        # that is NOT main_reaction_id and is NOT excluded, otherwise we double
        # up on the count/reaction shown in the UI.
        is_reaction_like_duplicate =
          reaction_users_counting_as_like.any? { |ru| ru.user_id == post_action.user_id }

        next true if is_reaction_like_duplicate

        # Also get rid of any PostAction records that match up to a ReactionUser
        # that is now the main_reaction_id and has historical data.
        is_previously_enabled_reaction =
          object
            .post_actions_with_reaction_users
            &.find do |pa|
              pa.id == post_action.id &&
                pa.reaction_user&.reaction&.reaction_value ==
                  DiscourseReactions::Reaction.main_reaction_id
            end
            .present?

        is_previously_enabled_reaction
      end
```

A bit of a micro-optimisation but you can avoid work if `is_reaction_like_duplicate` is already true.

tgxworld (approved):

plugin.rb
martin-brennan: Ah yeah good point, will change :+1:

PR #296, opened 2024-05-24
FIX: couldn't select text after dismissing a reaction

ZogStriP:
On mobile (touch device more accurately), if you were to open the reaction picker and then dismiss it by clicking outside, you would not be able to select text anymore.

That's because sometimes, the `touchEnd` event wouldn't trigger, and we would not remove the `discourse-reactions-no-select` class on the `html` element.

In order to fix this, I move the removal of the `discourse-reactions-no-select` class to the `collapseAllPanels` method which is always called when clicking outside.

Ref - https://meta.discourse.org/t/opening-and-closing-the-reactions-menu-without-picking-a-reaction-disables-highlighting/287064

NOTE: there are no tests, because I couldn't figure out a reliable enough way to test it.

### DIFF
--- a/assets/javascripts/discourse/widgets/discourse-reactions-actions.js
+++ b/assets/javascripts/discourse/widgets/discourse-reactions-actions.js
@@ -105,6 +105,7 @@ export default createWidget("discourse-reactions-actions", {
   defaultState() {
     return {
       reactionsPickerExpanded: false,
+      statePanelExpanded: false,
     };
   },
 
@@ -172,9 +173,11 @@ export default createWidget("discourse-reactions-actions", {
     this._validTouch = true;
 
     cancel(this._touchTimeout);
+
     if (this.capabilities.touch) {
-      const root = document.getElementsByTagName("html")[0];
-      root?.classList?.add("discourse-reactions-no-select");
+      document
+        .querySelector("html")
+        ?.classList?.toggle("discourse-reactions-no-select", true);
 
       this._touchStartAt = Date.now();
       this._touchTimeout = later(() => {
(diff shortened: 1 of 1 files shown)


assets/javascripts/discourse/widgets/discourse-reactions-actions.js
```diff
@@ -105,6 +105,7 @@ export default createWidget("discourse-reactions-actions", {
   defaultState() {
     return {
       reactionsPickerExpanded: false,
+      statePanelExpanded: false,
```
ZogStriP, line 108: Not part of the fix, but felt like it's better to have a default value for that one as well since we use it as part of the state.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js
```diff
@@ -172,9 +173,11 @@ export default createWidget("discourse-reactions-actions", {
     this._validTouch = true;
 
     cancel(this._touchTimeout);
+
     if (this.capabilities.touch) {
-      const root = document.getElementsByTagName("html")[0];
-      root?.classList?.add("discourse-reactions-no-select");
+      document
```
ZogStriP, line 178: Changed to use `toggle` instead of `add` to ensure we only ever have that class once.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js
```diff
@@ -199,9 +202,6 @@ export default createWidget("discourse-reactions-actions", {
       return;
     }
 
-    const root = document.getElementsByTagName("html")[0];
```
ZogStriP, line 202: This bit got moved somewhere called more reliably when the reaction panels are closed/collapsed.

assets/javascripts/discourse/widgets/discourse-reactions-actions.js
```diff
@@ -581,6 +581,9 @@ export default createWidget("discourse-reactions-actions", {
 
   collapseAllPanels() {
     cancel(this._collapseHandler);
+    document
```
ZogStriP, line 584: Felt like the safest spot to remove the class (again using `.toggle`)

awesomerobot (approved):

PR #312, opened 2024-10-11
PERF: Speed up `PostSerializer#reactions` by avoiding nested loops

tgxworld:
This change speeds up the `PostSerializer#reactions` method by
optimising the following line: `reaction_users_counting_as_like.any? { |ru| ru.user_id == post_action.user_id }`.
`Array#any?` has a worst case complexity of `O(N)` which is very slow
when the `reaction_users_counting_as_like` array has alot of records.
Since that line was called within the `object.post_actions.reject` loop,
we end up looping through the `reaction_users_counting_as_like` array N
times where N is the number of `object.post_actions` records.

The optimisation we are applying here is to change the
`reaction_users_counting_as_like` into a `Set` which only holds the
`user_id` of `ReactionUser` records. Look up for a `Set` is O(1) making
the method way more efficient.

Benchmarking this method locally with roughly 2000 `ReactionUser`
records and 8000 `PostAction` records reduces the runtime of the method
from roughly 5 seconds to 5ms.

### DIFF
--- a/plugin.rb
+++ b/plugin.rb
@@ -82,7 +82,7 @@ after_initialize do
 
   add_to_serializer(:post, :reactions) do
     reactions = []
-    reaction_users_counting_as_like = []
+    reaction_users_counting_as_like = Set.new
 
     object
       .emoji_reactions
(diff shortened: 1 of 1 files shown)


martin-brennan (approved):
Amazing :ok_hand:

plugin.rb
```diff
@@ -99,18 +99,16 @@ class Engine < ::Rails::Engine
...
 
-    reaction_users_counting_as_like.flatten!
-
     likes =
       object.post_actions.reject do |post_action|
         # Get rid of any PostAction records that match up to a ReactionUser
         # that is NOT main_reaction_id and is NOT excluded, otherwise we double
         # up on the count/reaction shown in the UI.
-        next true if reaction_users_counting_as_like.any? { |ru| ru.user_id == post_action.user_id }
+        next true if reaction_users_counting_as_like.include?(post_action.user_id)
```
tgxworld, line 111: `Set#include?` is O(1) complexity.

plugin.rb
```diff
@@ -99,18 +99,16 @@ class Engine < ::Rails::Engine
         if !DiscourseReactions::Reaction.reactions_excluded_from_like.include?(
              reaction.reaction_value,
            ) && reaction.reaction_value != DiscourseReactions::Reaction.main_reaction_id
-          reaction_users_counting_as_like << reaction.reaction_users
+          reaction_users_counting_as_like.merge(reaction.reaction_users.pluck(:user_id))
```
tgxworld, line 102: Using `pluck` so that we don't have to load `ReactionUser` records into memory when we are only using `ReactionUser#user_id` eventually.

PR #313, opened 2024-10-11
PERF: Speed up `PostSerializer#reactions` by avoiding nested loop take 2

tgxworld:
This is a follow-up to 7467349d6cd428f4e566cbbbfe3c5b8f37efdf99.

Within the `object.post_actions.reject` loop, we are running through
another loop given by `object.post_actions_with_reaction_users&.find`.
When the `object.post_actions` and
`object.post_actions_with_reaction_users` array is large, we end up
spending alot of time executing the loops.

This commit resolves the problem by reducing the number of records we
loop in `object.post_actions_with_reaction_users` by prefiltering the
records by `post_action.id` to remove alot of unnecessary iterations in
the loop.

Local benchmarks for the method for 8000 items in `object.post_actions` records and 2000
items in `post_actions_with_reaction_users` shows the runtime of the
method decreasing from roughly 7seconds to about 5ms.

### DIFF
--- a/lib/discourse_reactions/topic_view_serializer_extension.rb
+++ b/lib/discourse_reactions/topic_view_serializer_extension.rb
@@ -13,6 +13,11 @@ module DiscourseReactions::TopicViewSerializerExtension
         "post_actions.post_id IN (#{DiscourseReactions::PostActionExtension.post_action_with_reaction_user_sql})",
         valid_reactions: DiscourseReactions::Reaction.reactions_counting_as_like,
       )
+      .reduce({}) do |hash, post_action|
+        hash[post_action.post_id] ||= {}
+        hash[post_action.post_id][post_action.id] = post_action
+        hash
+      end
   end
 
   def posts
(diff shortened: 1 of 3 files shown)


lib/discourse_reactions/topic_view_serializer_extension.rb
```diff
@@ -29,8 +35,7 @@ def posts
         )
 
       posts.each do |post|
-        post.post_actions_with_reaction_users =
-          post_actions_with_reaction_users.select { |post_action| post_action.post_id == post.id }
```
tgxworld, line 33: This was looping through the whole array once per post which is inefficient. Instead, we reduce the array into a hash keyed by the `post_action.post_id` instead so that we can just do a O(1) hash lookup when setting `post.post_actions_with_reaction_users`.

plugin.rb
```diff
@@ -112,14 +112,10 @@ class Engine < ::Rails::Engine
...
-          &.find do |pa|
-            pa.id == post_action.id &&
-              pa.reaction_user&.reaction&.reaction_value ==
-                DiscourseReactions::Reaction.main_reaction_id
-          end
-          .present?
+        object.post_actions_with_reaction_users[post_action.id]
+          &.reaction_user
+          &.reaction
+          &.reaction_value == DiscourseReactions::Reaction.main_reaction_id
```
tgxworld, line 118: No more nested loop here and we can just do a O(1) hash lookup.

SamSaffron (approved):

PR #334, opened 2025-01-23
DEV: FA6 updates for icon setting description and badge fixture

tyb-talks:
Core badges migrations were already done in https://github.com/discourse/discourse/pull/29958.

For more detail on the FA6 upgrade, refer to the announcement at https://meta.discourse.org/t/-/325349.

### DIFF
--- a/db/fixtures/001_badges.rb
+++ b/db/fixtures/001_badges.rb
@@ -15,7 +15,7 @@ SQL
 
 Badge.seed(:name) do |b|
   b.name = "First Reaction"
-  b.default_icon = "smile"
+  b.default_icon = "face-smile"
   b.badge_type_id = BadgeType::Bronze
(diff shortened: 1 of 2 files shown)


db/fixtures/001_badges.rb
```diff
@@ -15,7 +15,7 @@
 
 Badge.seed(:name) do |b|
   b.name = "First Reaction"
-  b.default_icon = "smile"
+  b.default_icon = "face-smile"
```
Drenmi, line 18: As opposed to smiles that aren't on the face? 😛

Drenmi (approved):
🎸

db/fixtures/001_badges.rb
tyb-talks: ikr 🙃

PR #338, opened 2025-02-05
DEV: migrate smile to face-smile in badge icons

tyb-talks:
This ensures the change to update the deprecated icon name `smile` in the badge seed fixture is also applied to existing data in the DB.

### DIFF
--- a/db/migrate/20250205104150_remap_deprecated_icon_names_for_badge_fixtures.rb
+++ b/db/migrate/20250205104150_remap_deprecated_icon_names_for_badge_fixtures.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+class RemapDeprecatedIconNamesForBadgeFixtures < ActiveRecord::Migration[7.2]
+  def up
+    execute <<~SQL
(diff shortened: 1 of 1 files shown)


davidtaylorhq (approved):

db/migrate/20250205104150_remap_deprecated_icon_names_for_badge_fixtures.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+class RemapDeprecatedIconNamesForBadgeFixtures < ActiveRecord::Migration[7.2]
+  def up
+    execute <<~SQL
+      WITH remaps AS (
+        SELECT 'smile' AS from_icon, 'face-smile' AS to_icon
+      )
+      UPDATE badges
+      SET icon = remaps.to_icon
```
tyb-talks, line 9: 🤦 just realised I could have simplified this massively lol. with a single pair of values, we can just do a SET.. WHERE without the CTE. (reminder to self: don't do this)

PR #340, opened 2025-02-06
FIX: Add TopicViewPostsSerializerExtension to fix N1s

martin-brennan:
The following N1s occur when fetching posts in a topic
past the OP via the /t/:id/posts.json endpoint, which
passes in an array of post_ids:

```
SELECT "discourse_reactions_reactions".* FROM "discourse_reactions_reactions" WHERE "discourse_reactions_reactions"."post_id" = :id;
```

```
SELECT "post_actions".* FROM "post_actions" WHERE
"post_actions"."deleted_at" IS NULL AND "post_actions"."post_id" = :id;
```

```
SELECT union_subquery.post_id, COUNT(DISTINCT(union_subquery.user_id)) FROM (    SELECT user_id, post_id FROM post_actions      WHERE post_id IN (14755)        AND post_action_type_id = 2        AND deleted_at IS NULL  UNION ALL    SELECT discourse_reactions_reaction_users.user_id, posts.id from posts      LEFT JOIN discourse_reactions_reactions ON discourse_reactions_reactions.post_id = posts.id      LEFT JOIN discourse_reactions_reaction_users ON discourse_reactions_reaction_users.reaction_id = discourse_reactions_reactions.id      WHERE posts.id IN (14755)) AS union_subquery WHERE union_subquery.post_ID IS NOT NULL GROUP BY union_subquery.post_id;
```

This was happening because we hadn't added additional includes
and prefetching on the posts query for `TopicViewPostsSerializer`, like we were on
`TopicViewSerializerExtension`. This commit adds the same logic
to the `TopicViewPostsSerializer` via an extension to eliminate
the N1s.

### DIFF
--- a/lib/discourse_reactions/posts_reaction_loader.rb
+++ b/lib/discourse_reactions/posts_reaction_loader.rb
@@ -6,16 +6,15 @@ module DiscourseReactions::PostsReactionLoader
       posts = object.posts.includes(:post_actions, reactions: { reaction_users: :user })
       post_ids = posts.map(&:id).uniq
       posts_reaction_users_count = TopicViewSerializer.posts_reaction_users_count(post_ids)
-      posts.each { |post| post.reaction_users_count = posts_reaction_users_count[post.id].to_i }
-
       post_actions_with_reaction_users =
         DiscourseReactions::TopicViewSerializerExtension.load_post_action_reaction_users_for_posts(
           post_ids,
         )
-
       posts.each do |post|
+        post.reaction_users_count = posts_reaction_users_count[post.id].to_i
         post.post_actions_with_reaction_users = post_actions_with_reaction_users[post.id] || {}
       end
+
       object.instance_variable_set(:@posts, posts)
     end
   end
--- a/lib/discourse_reactions/posts_reaction_loader.rb
+++ b/lib/discourse_reactions/posts_reaction_loader.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module DiscourseReactions::PostsReactionLoader
+  def posts_with_reactions
+    if SiteSetting.discourse_reactions_enabled
+      posts = object.posts.includes(:post_actions, reactions: { reaction_users: :user })
+      post_ids = posts.map(&:id).uniq
+      posts_reaction_users_count = TopicViewSerializer.posts_reaction_users_count(post_ids)
+      posts.each { |post| post.reaction_users_count = posts_reaction_users_count[post.id].to_i }
+
+      post_actions_with_reaction_users =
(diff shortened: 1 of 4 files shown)


lib/discourse_reactions/topic_view_posts_serializer_extension.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module DiscourseReactions::TopicViewPostsSerializerExtension
+  # See also DiscourseReactions::TopicViewSerializerExtension
```
nattsw: This `def posts` looks like a full duplicate of the one in the other extension.

would it make sense to extract it into a `module PostsReactionLoader  def posts_with_reactions ...` and `include DiscourseReactions::PostsReactionLoader` then 
```
  def posts
    posts_with_reactions
  end
 ```
  ?

Comment (martin-brennan) [reply lib/discourse_reactions/topic_view_posts_serializer_extension.rb]: Yes I considered this but wasn't totally sure, but I think I will now :+1:

**Comment by martin-brennan** [comment]
@nattsw I made the change now

**Review by pento — APPROVED** [review]

File: lib/discourse_reactions/posts_reaction_loader.rb [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+      posts.each { |post| post.reaction_users_count = posts_reaction_users_count[post.id].to_i }
+
+      post_actions_with_reaction_users =
+        DiscourseReactions::TopicViewSerializerExtension.load_post_action_reaction_users_for_posts(
+          post_ids,
+        )
+
+      posts.each do |post|
+        post.post_actions_with_reaction_users = post_actions_with_reaction_users[post.id] || {}
+      end
```
Grubba27:
```suggestion
      posts = object.posts.includes(:post_actions, reactions: { reaction_users: :user })
      post_ids = posts.map(&:id).uniq

      posts_reaction_users_count = TopicViewSerializer.posts_reaction_users_count(post_ids)
      post_actions_with_reaction_users =
       DiscourseReactions::TopicViewSerializerExtension.load_post_action_reaction_users_for_posts(
         post_ids,
       )

      posts.each do |post|
        post.reaction_users_count = posts_reaction_users_count[post.id].to_i
        post.post_actions_with_reaction_users = post_actions_with_reaction_users[post.id] || {}
      end
```

🏌️: would this work? it would save a `post.each` run

Grubba27 (approved):

lib/discourse_reactions/posts_reaction_loader.rb
martin-brennan: Oh yeah good point @Grubba27 nice!

PR #342, opened 2025-02-07
DEV: Adds checks for 'prioritize_full_name_in_ux'

brrusselburg:
Equivalent core PR [here](https://github.com/discourse/discourse/pull/30583).

Ensures full names get passed through if the hidden `prioritize_full_name_in_ux` setting is `true`.

Before (with setting on):
Screenshot 2025-02-12 at 12 04 39 PM
Screenshot 2025-02-12 at 12 07 02 PM

After (with setting on):
Screenshot 2025-02-12 at 12 02 43 PM
Screenshot 2025-02-12 at 12 06 16 PM

### DIFF
--- a/app/services/discourse_reactions/reaction_notification.rb
+++ b/app/services/discourse_reactions/reaction_notification.rb
@@ -18,7 +18,7 @@ module DiscourseReactions
         return
       end
 
-      opts = { user_id: @user.id, display_username: @user.username }
+      opts = { user_id: @user.id, display_username: @user.username, display_name: @user.name }
 
       if @reaction.reaction_value == HEART_ICON_NAME
         opts[:custom_data] = { reaction_icon: @reaction.reaction_value }
@@ -52,7 +52,7 @@ module DiscourseReactions
         .joins(:users)
         .order("discourse_reactions_reactions.created_at DESC")
         .where("discourse_reactions_reactions.created_at > ?", 1.day.ago)
-        .pluck(:username, :reaction_value)
+        .pluck(:username, :name, :reaction_value)
     end
 
     def refresh_notification(read)
@@ -66,11 +66,13 @@ module DiscourseReactions
         count: remaining_data.length,
         username: remaining_data[0][0],
         display_username: remaining_data[0][0],
+        display_name: remaining_data[0][1],
       }
 
       data[:username2] = remaining_data[1][0] if remaining_data[1]
+      data[:name2] = remaining_data[1][1] if remaining_data[1]
 
-      if remaining_data.all? { |element| element[1] == HEART_ICON_NAME }
+      if remaining_data.all? { |element| element[2] == HEART_ICON_NAME }
         data[:reaction_icon] = HEART_ICON_NAME
       end
 
--- a/app/services/discourse_reactions/reaction_notification.rb
+++ b/app/services/discourse_reactions/reaction_notification.rb
@@ -70,7 +70,7 @@ module DiscourseReactions
       }
 
       data[:username2] = remaining_data[1][0] if remaining_data[1]
-      data[:name2] = remaining_data[1][1] if remaining_data[1] # something's not connecting correctly...
+      data[:name2] = remaining_data[1][1] if remaining_data[1]
 
       if remaining_data.all? { |element| element[2] == HEART_ICON_NAME }
         data[:reaction_icon] = HEART_ICON_NAME
--- a/app/services/discourse_reactions/reaction_notification.rb
+++ b/app/services/discourse_reactions/reaction_notification.rb
@@ -70,7 +70,7 @@ module DiscourseReactions
       }
 
       data[:username2] = remaining_data[1][0] if remaining_data[1]
-      data[:name2] = remaining_data[1][1] if remaining_data[1]
+      data[:name2] = remaining_data[1][1] if remaining_data[1] # something's not connecting correctly...
 
       if remaining_data.all? { |element| element[2] == HEART_ICON_NAME }
         data[:reaction_icon] = HEART_ICON_NAME
--- a/assets/javascripts/discourse/initializers/discourse-reactions.gjs
+++ b/assets/javascripts/discourse/initializers/discourse-reactions.gjs
@@ -132,7 +132,6 @@ function initializeDiscourseReactions(api) {
             ? this.notification.data.display_name || this.username
             : this.username;
 
-          // A single reaction works locally for me
           if (!count || count === 1 || !this.notification.data.username2) {
             return nameOrUsername;
           }
--- a/assets/javascripts/discourse/initializers/discourse-reactions.gjs
+++ b/assets/javascripts/discourse/initializers/discourse-reactions.gjs
@@ -135,6 +135,8 @@ function initializeDiscourseReactions(api) {
             return nameOrUsername;
           }
           if (count > 2) {
+            console.log("OVER TWO REACTIONS!!!!!!!!!");
+            console.log(nameOrUsername);
             return i18n("notifications.reaction_multiple_users", {
               username: nameOrUsername,
               count: count - 1,
--- a/assets/javascripts/discourse/initializers/discourse-reactions.gjs
+++ b/assets/javascripts/discourse/initializers/discourse-reactions.gjs
@@ -128,22 +128,43 @@ function initializeDiscourseReactions(api) {
 
         get label() {
           const count = this.notification.data.count;
+          const fullname = this.notification.data.original_name;
           const username = this.username;
 
+          // A single reaction works locally for me
           if (!count || count === 1 || !this.notification.data.username2) {
-            return username;
+            if (!this.siteSettings.prioritize_full_name_in_ux) {
+              return username;
+            } else {
+              return fullname || username;
+            }
           }
-
+          // the following two (react consolidation) are running into an issue
           if (count > 2) {
-            return i18n("notifications.reaction_multiple_users", {
-              username,
-              count: count - 1,
-            });
+            if (!this.siteSettings.prioritize_full_name_in_ux) {
+              return I18n.t("notifications.reaction_multiple_users", {
+                username,
+                count: count - 1,
+              });
+            } else {
+              return I18n.t("notifications.reaction_multiple_users", {
+                fullname,
+                count: count - 1,
+              });
+            }
           } else {
-            return i18n("notifications.reaction_2_users", {
-              username,
-              username2: formatUsername(this.notification.data.username2),
-            });
+            if (!this.siteSettings.prioritize_full_name_in_ux) {
+              console.log("HERE I AM");
+              return I18n.t("notifications.reaction_2_users", {
+                username,
+                username2: formatUsername(this.notification.data.username2),
+              });
+            } else {
+              return I18n.t("notifications.reaction_2_users", {
+                fullname,
+                fullname2: null, // can't get into this to figure out what this would be on 'this.notification.data.??'
+              });
+            }
           }
         }
 
--- a/assets/javascripts/discourse/initializers/discourse-reactions.gjs
+++ b/assets/javascripts/discourse/initializers/discourse-reactions.gjs
@@ -140,13 +140,12 @@ function initializeDiscourseReactions(api) {
               count: count - 1,
             });
           } else {
-            const name2OrUsername2 = this.siteSettings
-              .prioritize_full_name_in_ux
+            const nameOrUsername2 = this.siteSettings.prioritize_full_name_in_ux
               ? this.notification.data.name2
               : this.notification.data.username2;
             return i18n("notifications.reaction_2_users", {
               username: nameOrUsername,
-              username2: name2OrUsername2,
+              username2: nameOrUsername2,
             });
           }
         }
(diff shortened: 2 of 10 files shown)


martin-brennan (commented):
This does seem on the right track, did you want me to dig in to any of the failing stuff, or are you on top of it?

brrusselburg:
> This does seem on the right track, did you want me to dig in to any of the failing stuff, or are you on top of it?

Sweet, thanks! I poked around more today and found a couple things I forgot to update (mainly the yaml) but still ran into the same block. The reaction consolidation is happening in the ruby spec, so I assume the consolidation is getting lost in the js somewhere.

Screenshot 2025-02-10 at 5 11 00 PM

spec/system/reaction_notifications_spec.rb
```diff
@@ -63,4 +63,36 @@
...
+
+    it "renders reaction notifications with full names" do
+      visit("/")
+      user_menu.open
+
+      labels = page.all("#quick-access-all-notifications .notification.reaction .item-label")
+
+      expect(labels[0]).to have_text(
+        I18n.t(
+          "js.notifications.reaction_multiple_users.one",
```
brrusselburg, line 82: This fails saying it doesn't exist, but.... (see other comment)

config/locales/client.en.yml
```diff
@@ -28,12 +28,5 @@ en:
       reaction_multiple_users:
```
brrusselburg, line 28: ...this exists? I need a second pair of eyes

config/locales/client.en.yml
```diff
@@ -28,12 +28,5 @@ en:
...
         one: "%{username} and %{count} other"
```
brrusselburg, line 29: cc @jjaffeux -- I had the wrong line

assets/javascripts/discourse/initializers/discourse-reactions.gjs
```diff
@@ -128,21 +128,25 @@ function initializeDiscourseReactions(api) {
 
         get label() {
           const count = this.notification.data.count;
-          const username = this.username;
+          const nameOrUsername = this.siteSettings.prioritize_full_name_in_ux
+            ? this.notification.data.display_name
+            : this.notification.data.display_username;
```
OsamaSayegh:
```suggestion
            : this.username;
```

It currently uses `this.username` so we should keep it. They're practically the same, the only difference is that `this.username` has the `formatUsername` function applied to it whereas `this.notification.data.display_username` is the raw username.

test/javascripts/acceptance/discourse-reactions-notifications-test.js
```diff
@@ -241,3 +241,130 @@ acceptance("Discourse Reactions - Notifications", function (needs) {
     );
   });
 });
+
+acceptance(
+  "Discourse Reactions - Notifications | Full Name Setting On",
+  function (needs) {
+    needs.user({ redesigned_user_menu_enabled: true });
```
OsamaSayegh: This shouldn't be needed anymore. This was a feature flag to enable the notifications menu when it was still experimental, but that's not the case anymore and the feature flag is long gone.

spec/system/reaction_notifications_spec.rb
```diff
@@ -63,4 +63,33 @@
     )
     expect(labels[2]).to have_text(acting_user_1.username)
   end
+
+  context "when prioritize full names in ux site setting is on" do
+    before do
+      SiteSetting.prioritize_full_name_in_ux = true
+      SiteSetting.discourse_reactions_enabled = true
```
OsamaSayegh: I'd move this into a top level `before` block since every test here needs the plugin to be enabled.

test/javascripts/acceptance/discourse-reactions-notifications-test.js
```diff
@@ -241,3 +241,130 @@ acceptance("Discourse Reactions - Notifications", function (needs) {
...
+      assert.strictEqual(
+        notifications[1].textContent.replaceAll(/\s+/g, " ").trim(),
+        `${i18n("notifications.reaction_2_users", {
+          username: "Bruce Wayne I",
+          username2: "Brucer Wayner II",
+        })} Topic with 2 likes (total) from 2 users`,
+        "notification for 2 likes from 2 users has the right content"
+      );
+
+      // this one isn't receiving username value for the translation for some reason - solution should also fix the rails test error
```
OsamaSayegh: Leftover comment?

OsamaSayegh (commented):
Looks pretty good, just a few minor comments and it should be ready to go 🚀

spec/system/reaction_notifications_spec.rb
brrusselburg: I removed it, there's one at the top already 🧹

test/javascripts/acceptance/discourse-reactions-notifications-test.js
brrusselburg: Removed it but then the spec failed, so I put it back in

app/services/discourse_reactions/reaction_notification.rb
```diff
@@ -18,7 +18,7 @@ def create
         return
       end
 
-      opts = { user_id: @user.id, display_username: @user.username }
+      opts = { user_id: @user.id, display_username: @user.username, display_name: @user.name }
```
cvx, line 21: a general question for the entire PR: does it fallback to the username if `name` is empty?

brrusselburg: Just tested and it does _not_ 😭 thank you @CvX!

test/javascripts/acceptance/discourse-reactions-notifications-test.js
OsamaSayegh: Sorry I wasn't clear here — the `redesigned_user_menu_enabled` argument shouldn't be needed, but the `needs.user()` call needs to stay since it's what sets up the current user object for the test. Without `needs.user()`, we get no current user for the test and it fails because there's no notifications menu for anonymous users.

brrusselburg: OHH, I see! Thank you very much for clarifying :)

assets/javascripts/discourse/widgets/discourse-reactions-list-emoji.js
```diff
@@ -41,7 +41,12 @@ export default createWidget("discourse-reactions-list-emoji", {
     } else {
       users.slice(0, DISPLAY_MAX_USERS).forEach((user) => {
         let displayName;
-        if (this.siteSettings.prioritize_username_in_ux) {
+        if (
+          !this.siteSettings.prioritize_username_in_ux &&
+          this.siteSettings.prioritize_full_name_in_ux
+        ) {
+          displayName = user.name;
+        } else if (this.siteSettings.prioritize_username_in_ux) {
```
cvx: this doesn't fallback to username?

assets/javascripts/discourse/initializers/discourse-reactions.gjs
```diff
@@ -128,21 +128,26 @@ function initializeDiscourseReactions(api) {
 
         get label() {
           const count = this.notification.data.count;
-          const username = this.username;
+          const nameOrUsername = this.siteSettings.prioritize_full_name_in_ux
+            ? this.notification.data.display_name ||
+              this.notification.data.username
```
OsamaSayegh:
```suggestion
              this.username
```

assets/javascripts/discourse/initializers/discourse-reactions.gjs
```diff
@@ -128,21 +128,26 @@ function initializeDiscourseReactions(api) {
...
+            : this.username;
```
cvx: what's the diff between `this.notification.data.username` and `this.username`? and should both be used there? (depending on the site setting? 🤔)

assets/javascripts/discourse/initializers/discourse-reactions.gjs
```diff
@@ -128,21 +128,26 @@ function initializeDiscourseReactions(api) {
...
-
           if (count > 2) {
             return i18n("notifications.reaction_multiple_users", {
-              username,
+              username: nameOrUsername,
               count: count - 1,
             });
           } else {
+            const nameOrUsername2 = this.siteSettings.prioritize_full_name_in_ux
+              ? this.notification.data.name2 || this.notification.data.username2
```
OsamaSayegh:
```suggestion
              ? this.notification.data.name2 || formatUsername(this.notification.data.username2)
```

(or possibly extract `formatUsername(this.notification.data.username2)` into a variable and use it in the 2 spots in this ternary condition, up to you)

test/javascripts/acceptance/discourse-reactions-notifications-test.js
```diff
@@ -241,3 +241,129 @@ acceptance("Discourse Reactions - Notifications", function (needs) {
...
+                name2: "Roronoa Zoro",
+                count: 4,
+              },
+            },
+          ],
+        });
+      });
+    });
+
+    test("reaction notifications with full name site setting on", async (assert) => {
```
cvx: it's better to stick to the non-arrow functions as test methods (doesn't matter in this case, but if you were to use `this`, it wouldn't work)

```suggestion
    test("reaction notifications with full name site setting on", async function (assert) {
```

spec/system/reaction_notifications_spec.rb
```diff
@@ -63,4 +63,29 @@
     )
     expect(labels[2]).to have_text(acting_user_1.username)
   end
+
+  context "when prioritize full names in ux site setting is on" do
+    before { SiteSetting.prioritize_full_name_in_ux = true }
+
+    it "renders reaction notifications with full names" do
```
OsamaSayegh, line 81: Now that we've added the username fallback, I think it's worth adding a test case for when the full name setting is on but users have a null name to test that we actually fallback to the username.

OsamaSayegh (commented):
we're this 🤏 close to merging the PR, just a couple more things 🚀

test/javascripts/acceptance/discourse-reactions-notifications-test.js
```diff
@@ -241,3 +241,129 @@ acceptance("Discourse Reactions - Notifications", function (needs) {
...
+                original_post_type: 1,
+                original_username: "krus",
+                revision_number: null,
+                display_username: "krus",
+                display_name: "Brucer Wayner II",
+                reaction_icon: "heart",
+                previous_notification_id: 933,
+                count: 1,
+              },
+            },
```
cvx: if this notification isn't referenced in the test we can remove it imo

test/javascripts/acceptance/discourse-reactions-notifications-test.js
```diff
@@ -241,3 +241,129 @@ acceptance("Discourse Reactions - Notifications", function (needs) {
...
+      );
+
+      assert.strictEqual(
+        notifications[1].textContent.replaceAll(/\s+/g, " ").trim(),
+        `${i18n("notifications.reaction_2_users", {
+          username: "Bruce Wayne I",
+          username2: "Brucer Wayner II",
+        })} Topic with 2 likes (total) from 2 users`,
+        "notification for 2 likes from 2 users has the right content"
+      );
```
cvx: both assertions could be rewritten like this, avoiding the need for `queryAll` (I want to deprecate it at some point, it's mostly gone from core), manual reaching into `textContent`, regex replace, etc.

```suggestion
      assert
        .dom("li.notification.reaction:nth(1) a")
        .hasText(
          /Bruce Wayne I, Brucer Wayner II\s+Topic with 2 likes (total) from 2 users/,
        );
```

assets/javascripts/discourse/initializers/discourse-reactions.gjs
OsamaSayegh: `this.username` is a shorthand for `formatUsername(this.notification.data.username)` defined in `NotificationTypeBase`.

brrusselburg: turns out this was an error on my part, because on the notif object it's referred to as `display_username`. There isn't a `.username` on the notification object.

assets/javascripts/discourse/initializers/discourse-reactions.gjs
brrusselburg: Any reason this shouldn't be `this.notification.data.display_username`? Turns out `.username` doesn't exist on the notification object, so the way I had it wouldn't have worked anyway.

OsamaSayegh: Yes `this.notification.data.username` doesn't exist, however `this.username` is just a [getter](https://github.com/discourse/discourse/blob/f744852f105f2ac355af843b646fb2362cacc46d/app/assets/javascripts/discourse/app/lib/notification-types/base.js#L113-L115) that returns `this.notification.data.display_username` with `formatUsernaem()` applied on it. Is `this.username` not returning anything in your tests?

assets/javascripts/discourse/widgets/discourse-reactions-list-emoji.js
brrusselburg: oop, missed this one. thanks!

assets/javascripts/discourse/initializers/discourse-reactions.gjs
brrusselburg: Ahh gotcha thank you! Nah it worked :)

blake-discourse (approved):

PR #355, opened 2025-03-14
UX: Reactions user activity page shows names on mentions when  site setting is true

jdmartinez1062:
Before 
imagen

After

imagen


brrusselburg (approved):
(discussed the failure in chat, appears to be a flake)

